PR #42, opened 2022-05-23
INGM-154 ensure pytest works correctly

MartinALL91:


tests/test_communication.py
```diff
@@ -245,12 +245,15 @@ def test_set_register_wrong_access(motion_controller):
...
-def test_get_sdo_register(motion_controller, uid, index, subindex, dtype, value):
-    mc, alias = motion_controller
-    mc.communication.set_register(uid, value, servo=alias)
-    test_value = mc.communication.get_sdo_register(
-        index, subindex, dtype, servo=alias)
-    assert test_value == value
+def test_get_sdo_register(read_config, motion_controller, uid, index,
+                          subindex, dtype, value):
+    eoe_comm = read_config["soem"]["eoe_comm"]
+    if not eoe_comm:
```
GullinBustin, line 251: Maybe it's better if you write something like:
```
if eoe_comm:
    pytest.skip("Protocol does not match")
```

GullinBustin (changes requested):

GullinBustin (approved):

PR #53, opened 2022-09-07
INGM-205 all logs are changed as info logs

WoodElv:


GullinBustin:
Changes was done in the wrong file

PR #57, opened 2022-09-12
INGM-183 Get direct current and quadrature are added

WoodElv:


tests/test_motion.py
```diff
@@ -17,7 +18,9 @@
 VELOCITY_SET_POINT_REGISTER = "CL_VEL_SET_POINT_VALUE"
 ACTUAL_VELOCITY_REGISTER = "CL_VEL_FBK_VALUE"
 CURRENT_QUADRATURE_SET_POINT_REGISTER = "CL_CUR_Q_SET_POINT"
+ACTUAL_QUADRATURE_CURRENT_REGISTER = "CL_CUR_Q_VALUE"
```
MartinALL91, line 21: Maybe we could create a constants file where we could define these values and import them where needed, instead of creating the same variables on two different files.

ingeniamotion/motion.py
```diff
@@ -414,6 +416,36 @@ def get_actual_velocity(self, servo=DEFAULT_SERVO, axis=DEFAULT_AXIS):
...
+    def get_actual_direct_current(self, servo=DEFAULT_SERVO, axis=DEFAULT_AXIS):
+        """
+        Returns actual direct current register.
+
+        Args:
+            servo (str): servo alias to reference it. ``default`` by default.
+            axis (int): servo axis. ``1`` by default.
+
+        Returns:
+            int: actual direct current value
```
GullinBustin, line 428: The return type is not int, it is float

ingeniamotion/motion.py
```diff
@@ -414,6 +416,36 @@ def get_actual_velocity(self, servo=DEFAULT_SERVO, axis=DEFAULT_AXIS):
...
+    def get_actual_quadrature_current(self, servo=DEFAULT_SERVO, axis=DEFAULT_AXIS):
+        """
+        Returns actual quadrature current register.
+
+        Args:
+            servo (str): servo alias to reference it. ``default`` by default.
+            axis (int): servo axis. ``1`` by default.
+
+        Returns:
+            int: actual quadrature current value
```
GullinBustin, line 443: The return type is not int, it is float

GullinBustin (changes requested):

ingeniamotion/motion.py
```diff
@@ -414,6 +416,36 @@ def get_actual_velocity(self, servo=DEFAULT_SERVO, axis=DEFAULT_AXIS):
...
+            axis (int): servo axis. ``1`` by default.
+
+        Returns:
+            float: actual direct current value
+
+        """
+        return self.mc.communication.get_register(self.ACTUAL_DIRECT_CURRENT_REGISTER,
+                                                  servo=servo, axis=axis)
+
+    def get_actual_quadrature_current(self, servo=DEFAULT_SERVO, axis=DEFAULT_AXIS):
```
GullinBustin, line 434: IMHO, `get_actual_direct_current `and `get_actual_quadrature_current` should be changed to `get_actual_current_direct` and `get_actual_current_quadrature` to be coerent with `set_current_direct` and `set_current_quadrature`

GullinBustin (changes requested):

GullinBustin (approved):

MartinALL91 (approved):

PR #66, opened 2022-11-03
[INGM-197] Add exceptions at write_disturbance_data and map_registers

pzinemanas:
Two exceptions were added:

- write_disturbance_data: when the length of `registers `and `self.mapped_registers` do not coincide. We assume the registers are not correctly configured in this case.
- map_registers: if the `registers` list is empty

### DIFF
--- a/ingeniamotion/disturbance.py
+++ b/ingeniamotion/disturbance.py
@@ -120,7 +120,6 @@ class Disturbance:
 
         Raises:
             IMDisturbanceError: If the register list is empty.
-
             IMDisturbanceError: If the register is not allowed to be mapped as
                 a disturbance register.
         """
@@ -180,7 +179,6 @@ class Disturbance:
         Raises:
             IMDisturbanceError: If the registers are not correctly 
                 configured.
-
             IMDisturbanceError: If buffer size is not enough for all the
                 registers and samples.
         """
--- a/ingeniamotion/disturbance.py
+++ b/ingeniamotion/disturbance.py
@@ -119,11 +119,15 @@ class Disturbance:
             int: max number of samples
 
         Raises:
+            IMDisturbanceError: If the register list is empty.
+
             IMDisturbanceError: If the register is not allowed to be mapped as
                 a disturbance register.
         """
         if not isinstance(registers, list):
             registers = [registers]
+        if len(registers) == 0:
+            raise IMDisturbanceError("registers list is empty")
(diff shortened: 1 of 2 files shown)


WoodElv (approved):

ingeniamotion/disturbance.py
```diff
@@ -119,11 +119,15 @@ def map_registers(self, registers):
             int: max number of samples
 
         Raises:
+            IMDisturbanceError: If the register list is empty.
+
```
MartinALL91, line 123: I think that we should not add a blank line between each exception.

ingeniamotion/disturbance.py
```diff
@@ -174,11 +178,16 @@ def write_disturbance_data(self, registers_data):
                 as in :func:`map_registers`.
 
         Raises:
+            IMDisturbanceError: If the registers are not correctly 
+                configured.
+
```
MartinALL91, line 183: I think that we should not add a blank line between each exception.

tests/test_disturbance.py
```diff
@@ -130,11 +129,9 @@ def test_write_disturbance_data_buffer_exception(disturbance):
         disturbance.write_disturbance_data([0] * disturbance.max_sample_number)
 
 
-@pytest.mark.skip("Exception is not implemented yet")
 @pytest.mark.soem
 @pytest.mark.eoe
 def test_write_disturbance_data_not_configured(disturbance):
-    # TODO Add exception in function for this test case
```
MartinALL91, line 137: The idea of this test was to check if a user tries to write disturbance data and no registers are mapped as disturbance registers.

MartinALL91 (changes requested):

ingeniamotion/disturbance.py
```diff
@@ -119,11 +119,14 @@ def map_registers(self, registers):
...
 
         Raises:
+            IMDisturbanceError: If the register list is empty.
             IMDisturbanceError: If the register is not allowed to be mapped as
                 a disturbance register.
         """
         if not isinstance(registers, list):
             registers = [registers]
+        if len(registers) == 0:
+            raise IMDisturbanceError("registers list is empty")
```
GullinBustin, line 129: I'm not sure if the exception should be an IMDisturbanceError or a ValueError, because the parameter is empty and this value is not allowed. What do you think?

GullinBustin:
The issue will be reopened and will be done in the future

PR #67, opened 2022-11-03
[INGM-219] Call disturbance_remove_data when MonitoringVersion >= 3

pzinemanas:


WoodElv (approved):

ingeniamotion/capture.py
```diff
@@ -353,7 +353,7 @@ def disable_disturbance(self, servo=DEFAULT_SERVO, version=None):
             return self.disable_monitoring(servo=servo, version=version)
         drive = self.mc.servos[servo]
         drive.disturbance_disable()
-        if version < MonitoringVersion.MONITORING_V3:
+        if version >= MonitoringVersion.MONITORING_V3:
```
MartinALL91, line 356: I would simplify the function like so:
```
if version is None:
    version = self._check_version(servo)
if not self.is_disturbance_enabled(servo, version):
    return
if version < MonitoringVersion.MONITORING_V3:
    return self.disable_monitoring(servo=servo, version=version)
drive = self.mc.servos[servo]
drive.disturbance_disable()
return drive.disturbance_remove_data()
```
What do you think?

MartinALL91 (approved):

GullinBustin (approved):

PR #72, opened 2022-12-01
INGM-260 Launch ingeniamotion tests

MartinALL91:
Update Jenkinsfile and config.json in order to run tests via Jenkins.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,17 +16,30 @@ pipeline {
                 }
             }
             stages {
+                stage('Clone repository') {
+                    steps {
+                        bat """
+                            cd C:\\Users\\ContainerAdministrator
+                            git clone https://github.com/ingeniamc/ingeniamotion.git
+                            cd ingeniamotion
+                            git checkout ${env.GIT_COMMIT}
+                        """
+                    }
+                 }
                 stage('Install deps') {
                     steps {
                         bat '''
+                            cd C:\\Users\\ContainerAdministrator\\ingeniamotion
                             python -m venv venv
                             venv\\Scripts\\python.exe -m pip install -r requirements\\dev-requirements.txt
+                            venv\\Scripts\\python.exe -m pip install -e .
                         '''
                     }
                 }
                 stage('Build wheels') {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -1,136 +1,181 @@
...
-            stage('Save test results') {
-                archiveArtifacts artifacts: '*.html'
+            stages {
+                stage('Clone repository') {
+                    steps {
+                        bat """
+                            cd C:\\Users\\ContainerAdministrator
+                            git clone https://github.com/ingeniamc/ingeniamotion.git
+                            cd ingeniamotion
+                            git checkout ${env.GIT_COMMIT}
```
GullinBustin, line 25: Why did you clone the repository?
why don't you use "checkout scm"?

Jenkinsfile
```diff
@@ -1,136 +1,181 @@
...
+                        """
+                    }
+                 }
+                stage('Install deps') {
+                    steps {
+                        bat '''
+                            cd C:\\Users\\ContainerAdministrator\\ingeniamotion
```
GullinBustin, line 32: Why do you go to the repository directory at each step? The workspace should already be in the correct directory

GullinBustin (changes requested):

WoodElv (approved):

Jenkinsfile
MartinALL91: Because If we don't clone the repository within the container the environment does not find the library as installed.

Jenkinsfile
MartinALL91: Because we are working with the cloned repository within the container, the workspace is located elsewhere.

MartinALL91: Check the output of Jenkins build # 17 of the branch test-INGM-260.

Jenkinsfile
GullinBustin: It will be solved in another issue

GullinBustin (approved):

pzinemanas (approved):

PR #74, opened 2023-01-03
INGM-261 Load configuration if a test fails

pzinemanas:
### Description

This PR adds a fixture to load configuration if a test fails. Based on [pytest docs](https://docs.pytest.org/en/latest/example/simple.html#making-test-result-information-available-in-fixtures).

Fixes # INGM-261

### Type of change

- [x] Add fixture.
- [x] Upgrade pytest to 7.0.1 (still compatible with py 3.6.0).

### Tests
- [x] It was verified that the fixture was called if a test fails during setup or calling.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,13 @@
 import json
 import pytest
+from typing import Dict
 
 from ingeniamotion.enums import CAN_BAUDRATE, CAN_DEVICE, SensorType
 from ingeniamotion import MotionController
 
 ALLOW_PROTOCOLS = ["eoe", "soem", "canopen"]
+
+test_report_key = pytest.StashKey[Dict[str, pytest.CollectReport]]()
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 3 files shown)


MartinALL91 (approved):

tests/conftest.py
```diff
@@ -133,3 +136,24 @@ def clean_and_restore_feedbacks(motion_controller):
...
+
+
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    # execute all other hooks to obtain the report object
+    outcome = yield
+    rep = outcome.get_result()
+
+    # store test results for each phase of a call, which can be "setup", "call", "teardown"
+    item.stash.setdefault(test_report_key, {})[rep.when] = rep
```
GullinBustin, line 148: Why you use `item.stash.setdefault(test_report_key, {})[rep.when] = rep` instead `item.stash[test_report_key] = {rep.when: rep}`?

tests/conftest.py
```diff
@@ -133,3 +136,24 @@ def clean_and_restore_feedbacks(motion_controller):
...
+
+
+@pytest.fixture(scope="function", autouse=True)
+def load_configuration_if_test_fails(request, motion_controller, read_config):
+    mc, alias = motion_controller
+    yield
+
+    report = request.node.stash[test_report_key]
```
GullinBustin, line 156: Could you let me know why you use request.node.stash instead item.stash? I don't know the difference but in documentation the second case is used.
https://docs.pytest.org/en/7.1.x/how-to/writing_hook_functions.html#storing-data-on-items-across-hook-functions

tests/conftest.py
pzinemanas: I just followed this code: https://docs.pytest.org/en/latest/example/simple.html#making-test-result-information-available-in-fixtures

tests/conftest.py
pzinemanas: I'm following the [pytest example code](https://docs.pytest.org/en/latest/example/simple.html#making-test-result-information-available-in-fixtures). I'm not sure about the difference between those two.

WoodElv (approved):

GullinBustin (approved):

PR #85, opened 2023-01-09
INGM-249 Improve tests of the homing module

pzinemanas:
### Description

This PR improves the tests of the homing module.

Fixes # INGM-249

### Type of change

- [x] Calculate the mean position instead of using one sample.

### Tests
- [x] Run tests and obtain similar results.


tests/test_homing.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 
 from ingeniamotion.enums import HomingMode, SensorType, OperationMode
+from .conftest import mean_actual_velocity_position
```
MartinALL91, line 6: Maybe you should use an absolute import for coherence's sake.

MartinALL91 (approved):

WoodElv (approved):

GullinBustin (approved):

PR #89, opened 2023-01-12
INGM-266 Improve load_FWs script using canopen protocol

pzinemanas:
### Description

This PR improves the script to load FW using canopen.

Fixes # INGM-266

### Type of change

- [x] Add attempts when loading FW.
- [x] Verify the drive responds after flashing.

### Tests
- [x] Run the script locally and in the CI setup.

### DIFF
--- a/tests/load_FWs.py
+++ b/tests/load_FWs.py
@@ -55,12 +55,11 @@ def load_can(drive_conf, mc):
             mc.communication.load_firmware_canopen(drive_conf["fw_file"])
 
             # Reaching this means that FW was correctly flashed
-            time.sleep(SLEEP_TIME_AFTER_ATTEMP)
             break
 
         except ILFirmwareLoadError as e:
             logger.error(f"CAN boot error: {e}")
-            raise e
+            time.sleep(SLEEP_TIME_AFTER_ATTEMP)
 
         finally:
             mc.communication.disconnect()
--- a/tests/load_FWs.py
+++ b/tests/load_FWs.py
@@ -14,7 +14,7 @@ from ingeniamotion.enums import CAN_BAUDRATE, CAN_DEVICE
 from ingenialink.exceptions import ILError, ILFirmwareLoadError
 
 logger = ingenialogger.get_logger("load_FWs")
-ingenialogger.configure_logger()
+ingenialogger.configure_logger(level=ingenialogger.LoggingLevel.INFO)
 dirname = os.path.dirname(__file__)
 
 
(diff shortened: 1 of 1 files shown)


GullinBustin (approved):

tests/load_FWs.py
```diff
@@ -25,35 +25,98 @@ def setup_command():
...
+    BL_NUM_OF_REATTEMPTS = 2
+
+    # Timings, in seconds
+    SLEEP_TIME_AFTER_ATTEMP = 5.0
+    SLEEP_TIME_AFTER_BL = 5.0
+    TIMEOUT_NEW_FW_DETECT = 30.0
+    SLEEP_TIME_NEW_FW_DETECT = 5.0
+
+    for attempt in range(BL_NUM_OF_REATTEMPTS):
+        logger.info("CAN boot attempt {} of {}".format(attempt + 1, BL_NUM_OF_REATTEMPTS))
```
MartinALL91, line 38: Use f-string

tests/load_FWs.py
```diff
@@ -25,35 +25,98 @@ def setup_command():
...
+        try:
```
MartinALL91, line 39: I think this try should only wrap up the `load_firmware_canopen` function.

tests/load_FWs.py
```diff
@@ -25,35 +25,98 @@ def setup_command():
...
+            time.sleep(SLEEP_TIME_AFTER_ATTEMP)
+            break
+
+        except ILFirmwareLoadError as e:
+            # Disconnect CAN network
+            mc.communication.disconnect()
+
+            if str(e) != "Could not recover drive":
+                # TODO Fix canopen FW loader and remove this if
+                err_str = "CAN boot error: {}".format(e)
```
MartinALL91, line 70: Use f-string

tests/load_FWs.py
```diff
@@ -25,35 +25,98 @@ def setup_command():
...
+                logger.error(err_str)
+                time.sleep(SLEEP_TIME_AFTER_ATTEMP)
```
MartinALL91, line 72: I don't think this sleep is necessary.

tests/load_FWs.py
```diff
@@ -25,35 +25,98 @@ def setup_command():
...
+                raise e
+            else:
+                err_str = None
+                logger.warning(f"Exception '{e}' has raised, but it is ignored.")
+                break
+
+    if err_str is None:
```
MartinALL91, line 79: I think the `err_str` is not needed, if you reach this line it means the firmware was loaded successfully

tests/load_FWs.py
```diff
@@ -25,35 +25,98 @@ def setup_command():
...
+            "FW updated. %s, node: %d, baudrate: %d, channel: %d",
+            drive_conf["device"],
+            drive_conf["node_id"],
+            drive_conf["baudrate"],
+            drive_conf["channel"],
+        )
+    else:
+        raise Exception(err_str)
+
+    logger.info("Waiting {} seconds for trying to connect".format(SLEEP_TIME_AFTER_BL))
```
MartinALL91, line 90: Use f-string

tests/load_FWs.py
```diff
@@ -25,35 +25,98 @@ def setup_command():
...
+            # When cannot connect
+            pass
+
+        mc.communication.disconnect()
+
+        if not detected:
+            time.sleep(SLEEP_TIME_NEW_FW_DETECT)
+
+    if not detected:
+        raise Exception("New FW not detected")
```
MartinALL91, line 118: Maybe we should use the `ILFirmwareLoadError` exception.

tests/load_FWs.py
```diff
@@ -25,35 +25,98 @@ def setup_command():
...
+            detected = True
+            logger.info("New FW detected after: {:.1f} s".format(time.perf_counter() - ini_time))
+        except Exception as e:
+            # When cannot connect
+            pass
+
+        mc.communication.disconnect()
+
+        if not detected:
+            time.sleep(SLEEP_TIME_NEW_FW_DETECT)
```
MartinALL91, line 115: Maybe we could move this sleep to the top of the while loop and remove the if statement.

MartinALL91 (approved):

tests/load_FWs.py
pzinemanas: I thought that script-level exceptions should be general `Exceptions` and let `ILFirmwareLoadError` for the IL-related exceptions.

tests/load_FWs.py
```diff
@@ -25,35 +25,89 @@ def setup_command():
...
+
+            mc.communication.disconnect()
+
+            # Reaching this means that FW was correctly flashed
+            time.sleep(SLEEP_TIME_AFTER_ATTEMP)
+            break
+
+        except ILFirmwareLoadError as e:
+            # Disconnect CAN network
+            mc.communication.disconnect()
```
MartinALL91, line 65: You can use the `finally` clause for this command and remove it from here and from line 57.

tests/load_FWs.py
```diff
@@ -25,35 +25,80 @@ def setup_command():
...
+                channel=drive_conf["channel"],
+            )
+            # Reaching this point means we are connected
+            detected = True
+            logger.info("New FW detected after: {:.1f} s".format(time.perf_counter() - ini_time))
+            mc.communication.disconnect()
+        except Exception as e:
+            # When cannot connect
+            time.sleep(SLEEP_TIME_NEW_FW_DETECT)
+            pass
```
MartinALL91, line 98: I think this line is not necessary

WoodElv (approved):

PR #125, opened 2023-11-06
INGM-337: run only smoke test in pr jobs

WoodElv:
### Description

Issues documentation.

Fixes # (issue)

### Type of change

- [x] Tests that need a setup and takes 1 second (approx.) as maximum are selected as smoke tests.
- [x] Having the option to run only smoke or all tests.
- [x] Running only smoke tests for develop or master branches is not allowed.
- [x] Running all implemented tests for develop or master branches is mandatory.

### Documentation

### Code formatting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 - Feedbacks functions to set/get feedback polarity
 - Add in configuration set_velocity_pid, set_position_pid and get_rated_current
 - Add functions to connect to and scan EtherCAT devices using CoE (SDOs).
+- Running Smoke and All tests stages for the Jenkins pipeline.
+- Option to choose running only smoke and no connection tests, or all tests.
 
 ### Fixed
 - check_motor_disabled decorator does not work with positional arguments
@@ -26,6 +28,7 @@
 ### Changed
 - Use ingenialink to get the drive's encoded image from the dictionary.
 - Update subscribe and unsubscribe to network status functions.
+- Smoke tests are defined as tests that need a setup but takes 1 second (approx.) as maximum.
 
 ## [0.6.3] - 2023-10-11
 ### Fixed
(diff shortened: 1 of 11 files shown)


Jenkinsfile
```diff
@@ -133,8 +139,8 @@ pipeline {
                 stage('Run EtherCAT tests') {
```
GullinBustin, line 139: Why you don't check params.TESTS here?

tests/test_configuration.py
```diff
@@ -597,7 +596,6 @@ def test_get_set_rated_current(motion_controller):
 
 
 @pytest.mark.develop
-@pytest.mark.smoke
```
GullinBustin, line 600: Why this is not smoke?

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
 - Feedbacks functions to set/get feedback polarity
 - Add in configuration set_velocity_pid, set_position_pid and get_rated_current
 - Add functions to connect to and scan EtherCAT devices using CoE (SDOs).
+- Running Smoke and All tests stages for the Jenkins pipeline.
+- Option to choose running only smoke and no connection tests, or all tests.
```
GullinBustin, line 21: I don't think these changes should be in the changelog. They are not library changes, they are changes in the process.
What do you think?

GullinBustin (changes requested):

CHANGELOG.md
WoodElv: But this PR adds the option to choose what the developer wants. I had to add a parameter for that in the Jenkinsfile. And the Jenkinsfile is inside the library, isn't it?

tests/test_configuration.py
WoodElv: Theoretically because it takes more than 1 second. I'll check it again, just in case.

CHANGELOG.md
GullinBustin: It refers to the development, not to the use.

MartinALL91: I also think that this change should not be in the CHANGELOG

MartinALL91 (approved):

GullinBustin (approved):

PR #126, opened 2023-11-08
INGM-348: existence checking for some registers.

WoodElv:
### Description

There are some registers that they are not in dictionaries. This PR adds a checking of their existence to avoid an error running test. 

Fixes # (INGM-348)

### Type of change

- [x] Existence checking for some registers to run a Feedbacks test.

### Tests

### Documentation

### Code formatting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,14 @@
 - Feedbacks functions to set/get feedback polarity
 - Add in configuration set_velocity_pid, set_position_pid and get_rated_current
 - Add functions to connect to and scan EtherCAT devices using CoE (SDOs).
+- Optional registers for Wizard tests.
+- All registers are initialized for no_connection unit tests.
 
 ### Fixed
 - check_motor_disabled decorator does not work with positional arguments
 - Adapt get_encoded_image_from_dictionary for COM-KIT
 - Feedback test output when symmetry and resolution errors occurred simultaneously.
+- No error because of positioning registers are missing in the dictionary.
 
 ### Changed
 - Use ingenialink to get the drive's encoded image from the dictionary.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,14 +17,12 @@
 - Feedbacks functions to set/get feedback polarity
 - Add in configuration set_velocity_pid, set_position_pid and get_rated_current
 - Add functions to connect to and scan EtherCAT devices using CoE (SDOs).
-- Optional registers for Wizard tests.
-- All registers are initialized for no_connection unit tests.
+- Optional backup registers for Wizard tests.
 
 ### Fixed
 - check_motor_disabled decorator does not work with positional arguments
 - Adapt get_encoded_image_from_dictionary for COM-KIT
 - Feedback test output when symmetry and resolution errors occurred simultaneously.
-- No error because of positioning registers are missing in the dictionary.
 
 ### Changed
 - Use ingenialink to get the drive's encoded image from the dictionary.
--- a/ingeniamotion/wizard_tests/base_test.py
+++ b/ingeniamotion/wizard_tests/base_test.py
@@ -1,16 +1,16 @@
-from typing import TYPE_CHECKING, Any, Dict, List, Union, Optional
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 import ingenialogger
-
-from abc import ABC, abstractmethod
 from ingenialink.exceptions import ILError
 
 from ingeniamotion.exceptions import IMRegisterNotExist, IMRegisterWrongAccess
 from ingeniamotion.metaclass import DEFAULT_SERVO
-from ingeniamotion.wizard_tests.stoppable import Stoppable, StopException
+from ingeniamotion.wizard_tests.stoppable import StopException, Stoppable
 
 if TYPE_CHECKING:
     from ingeniamotion import MotionController
+
 from ingeniamotion.enums import SeverityLevel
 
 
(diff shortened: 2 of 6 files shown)


ingeniamotion/wizard_tests/feedbacks_tests/feedback_test.py
```diff
@@ -79,9 +79,6 @@ class Polarity(IntEnum):
         "COMMU_ANGLE_REF_SENSOR",
         "CL_VEL_FBK_FILTER1_TYPE",
         "CL_VEL_FBK_FILTER1_FREQ",
-        POSITIONING_OPTION_CODE_REGISTER,
-        MAX_POSITION_RANGE_LIMIT_REGISTER,
-        MIN_POSITION_RANGE_LIMIT_REGISTER,
```
GullinBustin, line 84: Why do you remove that? If the backup register does not exist simply ignore it.

ingeniamotion/wizard_tests/feedbacks_tests/feedback_test.py
```diff
@@ -217,15 +231,24 @@ def setup(self) -> None:
...
+        ):
+            self.mc.communication.set_register(
+                self.MIN_POSITION_RANGE_LIMIT_REGISTER, 0, servo=self.servo, axis=self.axis
+            )
+        if self.mc.info.register_exists(
+            self.MAX_POSITION_RANGE_LIMIT_REGISTER, servo=self.servo, axis=self.axis
+        ):
+            self.mc.communication.set_register(
+                self.MAX_POSITION_RANGE_LIMIT_REGISTER, 0, servo=self.servo, axis=self.axis
+            )
```
GullinBustin, line 251: The setup function is getting bigger. Move this logic to a function.

ingeniamotion/wizard_tests/feedbacks_tests/feedback_test.py
```diff
@@ -100,9 +97,26 @@ def __init__(self, mc: "MotionController", servo: str, axis: int) -> None:
         self.pos_vel_same_feedback = False
         self.resolution_multiplier = 1.0
         self.test_frequency = self.TEST_FREQUENCY
+        self.add_registers_to_backup()
         self.backup_registers_names = self.BACKUP_REGISTERS
         self.suggested_registers = {}
 
+    def add_registers_to_backup(self) -> None:
```
GullinBustin, line 104: This function is not necessary. See the comment above.

GullinBustin (changes requested):

ingeniamotion/wizard_tests/feedbacks_tests/feedback_test.py
WoodElv: This function is used to ignore registers that are not needed in an active way. I think it's clearer from conceptual point of view in this way. What do you think?

WoodElv: The thing is sometimes there are drives that they don't need some registers to run a test. That means these registers are not in the dictionary, thus, a log warning will be raised and it seems something is going badly.

In order to differentiate what is a mistake and what is not a mistake, we conclude to add optional registers to append them to the backup registers list for these tests.
If an optional register exists means it's important for a proper test, instead the test can be run without any problem.

This logic will be moved to the base test class, and the optional registers will be selected in the child classes.

CHANGELOG.md
```diff
@@ -17,11 +17,14 @@
...
 - Add in configuration set_velocity_pid, set_position_pid and get_rated_current
 - Add functions to connect to and scan EtherCAT devices using CoE (SDOs).
+- Optional registers for Wizard tests.
+- All registers are initialized for no_connection unit tests.
 
 ### Fixed
 - check_motor_disabled decorator does not work with positional arguments
 - Adapt get_encoded_image_from_dictionary for COM-KIT
 - Feedback test output when symmetry and resolution errors occurred simultaneously.
+- No error because of positioning registers are missing in the dictionary.
```
GullinBustin, line 27: This point can be confusing for external users.

- `No error because of positioning registers are missing in the dictionary.` This error does not happen in previous versions, so, I'm not sure that it should written in the changelog.
- `All registers are initialized for no_connection unit tests` This talks about tests, not about library behavior. This should be added in PR description.
- `Optional registers for Wizard tests.` That's ok, but maybe you mention the backup registers because is not easy to know what you mean by "Optional registers"

ingeniamotion/wizard_tests/base_test.py
```diff
@@ -42,6 +43,13 @@ def save_backup_registers(self) -> None:
             except IMRegisterNotExist as e:
                 self.logger.warning(e, axis=self.axis)
 
+        if not isinstance(self.optional_backup_registers_names, List):
```
GullinBustin, line 46: Why do you do that? Can optional_backup_registers_names not be a list?

GullinBustin (changes requested):

ingeniamotion/wizard_tests/base_test.py
pzinemanas: Moreover, I think `List` (with uppercase) should be used only for typing.

pzinemanas (approved):

CHANGELOG.md
```diff
@@ -17,11 +17,14 @@
 - Feedbacks functions to set/get feedback polarity
 - Add in configuration set_velocity_pid, set_position_pid and get_rated_current
 - Add functions to connect to and scan EtherCAT devices using CoE (SDOs).
+- Optional registers for Wizard tests.
```
MartinALL91, line 20: Optional backup registers, maybe?

MartinALL91 (approved):

GullinBustin (approved):

PR #144, opened 2024-02-01
INGM-386 Use ingenialink to load the FW

pzinemanas:
### Description

This PR changes the load FW script to use ingenialink instead of ingeniamotion.

Fixes # INGM-386 

### Type of change

- [x] Use ingenialink instead of ingeniamotion to load the FW.

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### DIFF
--- a/tests/load_FWs.py
+++ b/tests/load_FWs.py
@@ -1,17 +1,15 @@
-import sys
+import argparse
+import json
 import os
 import time
-import json
-import argparse
+from functools import partial
+
 import ingenialogger
-from ping3 import ping
-
-sys.path.append("./")
-
-from ingeniamotion import MotionController
-from ingeniamotion.exceptions import IMException
-from ingeniamotion.enums import CAN_BAUDRATE, CAN_DEVICE
+from ingenialink.canopen.network import CAN_BAUDRATE, CAN_DEVICE, CanopenNetwork
+from ingenialink.ethercat.network import EthercatNetwork
+from ingenialink.ethernet.network import EthernetNetwork
 from ingenialink.exceptions import ILError, ILFirmwareLoadError
+from ping3 import ping
 
 logger = ingenialogger.get_logger("load_FWs")
 ingenialogger.configure_logger(level=ingenialogger.LoggingLevel.INFO)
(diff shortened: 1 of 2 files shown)


tests/load_FWs.py
```diff
@@ -87,36 +93,40 @@ def load_can(drive_conf, mc):
...
+                f"New FW detected ({fw_version}) after: {time.perf_counter() - ini_time:.1f} s"
+            )
+            net.disconnect_from_slave(servo)
         except Exception as e:
             # When cannot connect
             time.sleep(SLEEP_TIME_NEW_FW_DETECT)
 
     if not detected:
-        raise Exception("New FW not detected")
+        raise Exception("Could not connect to the drive. FW loading might have failed.")
```
WoodElv, line 109: I think you could use a more specific Exception. For instance ILError. What do you think?

tests/load_FWs.py
```diff
@@ -87,36 +93,40 @@ def load_can(drive_conf, mc):
...
-            logger.info("New FW detected after: {:.1f} s".format(time.perf_counter() - ini_time))
-            mc.communication.disconnect()
+            fw_version = servo.info["firmware_version"]
+            logger.info(
+                f"New FW detected ({fw_version}) after: {time.perf_counter() - ini_time:.1f} s"
+            )
+            net.disconnect_from_slave(servo)
         except Exception as e:
             # When cannot connect
             time.sleep(SLEEP_TIME_NEW_FW_DETECT)
```
WoodElv, line 106: Why this sleep is needed?

tests/load_FWs.py
WoodElv: In fact, if you don't use ILError exception, this exception won't be handled in the main function, right? Or is your intention because Jenkins will detect it, and the job building will stop?

WoodElv (changes requested):

tests/load_FWs.py
pzinemanas: I just copied the code from ingenialink: I think the idea is what you mentioned, if the exception is raised, the Jenkins job is stopped.

tests/load_FWs.py
pzinemanas: Not sure. I guess it waits to let the drive start the program.

WoodElv (approved):

GullinBustin (approved):

PR #145, opened 2024-02-02
INGM-362 Add tox to test all Python versions

pzinemanas:
### Description

This PR adds tox to test all Python versions and changes the Jenkins pipeline

Fixes # INGM-362

### Type of change

- [x] Add tox to test all Python versions.
- [x] Change Jenkins pipeline.

### Tests
- [x] Run smoke tests (job # 1)
- [x] Run all tests (job # 8)

### Documentation

### Code formatting


WoodElv (approved):

tox.ini
```diff
@@ -0,0 +1,80 @@
+[tox]
+requires =
+    tox>=4
+env_list = coverage, build, firmware, docs, format, type, py{39,310,311,312}
+
+[testenv]
+description = run unit tests
+deps = 
+    -r{toxinidir}/requirements.txt
+    pytest==7.0.1
```
MartinALL91, line 10: I think we should create a requirements file for these dependencies.

MartinALL91: To avoid defining dependencies in two places (tox.ini and requirements.txt). If we define the testing and dev dependencies in a txt file, we could simply use them here.

pzinemanas: Well, I wanted to remove the requirements.txt as well. I kept it because sometimes we want to point to a specific branch of ingenialink as it is now in the develop branch.

GullinBustin: requirements.txt has no sense anymore. For install ingenialink develop branch we can add a file to these cases. But usually it will not be necessary

pzinemanas: Ok, I added a custom_requirements file to keep all requirements in the tox.ini file.

pzinemanas: @WoodElv found a way to avoid using a custom_requirements file: [69d99a4](https://github.com/ingeniamc/ingeniamotion/pull/145/commits/69d99a49db47db6a0f88eba31de8e21befd6c815)

GullinBustin (approved):

PR #147, opened 2024-02-12
INGM-388 Fix Jenkins pipeline

pzinemanas:
This PR fixes the Jenkins pipeline.

Doc: https://novantamotion.atlassian.net/wiki/spaces/SKB/pages/553189390/Locking+machines+with+real+drives

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,9 @@
 def SW_NODE = "windows-slave"
+def ECAT_NODE = "ecat-test"
+def ECAT_NODE_LOCK = "test_execution_lock_ecat"
+def CAN_NODE = "canopen-test"
+def CAN_NODE_LOCK = "test_execution_lock_can"
+
 def DOCKER_IMAGE = "ingeniacontainers.azurecr.io/win-python-builder:1.4"
 
 def DEFAULT_TOX_PYTHON_VERSION = "39"
@@ -6,22 +11,6 @@ def DEFAULT_PYTHON_VERSION = "3.9"
 def TOX_VERSION = "4.12.1"
 
 def SMOKE_TESTS_FLAG = ""
-
-def getAgentForProtocol(String protocol) {
-    if (protocol == 'soem') {
-        return "ecat-test"
-    } else {
-        return "canopen-test"
-    }
-}
-
-def getLockForProtocol(String protocol) {
-    if (protocol == 'soem') {
-        return "test_execution_lock_ecat"
-    } else {
-        return "test_execution_lock_can"
-    }
-}
 
 def dockerInstallTox = {
     bat """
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -112,136 +101,223 @@ pipeline {
...
-                            always {
-                                bat """
-                                    move .coverage .coverage_${PROTOCOL}
-                                """    
-                                junit "pytest_reports\\*.xml"
-                                archiveArtifacts artifacts: 'pytest_reports\\*.xml'
-                                stash includes: ".coverage_${PROTOCOL}", name: "coverage_report_${PROTOCOL}"
+                            stage('Run tests') {
+                                options {
+                                    lock('sequential-matrix')
```
polfeliu, line 290: what is this?

Jenkinsfile
```diff
@@ -112,136 +101,223 @@ pipeline {
...
+                lock(ECAT_NODE_LOCK)
+            }
+            agent {
+                label ECAT_NODE
+            }
+            stages {
+                stage("Checkout") {
+                    steps {
+                        checkout scm
+                        script {installTox ()}
```
polfeliu, line 115: I don't understand why we are doing all these changes here and locking the machine also during package installation instead of fixing ingenialink

polfeliu (dismissed):
missing doc

WoodElv (approved):

Jenkinsfile
pzinemanas: This is to force that the matrix is running sequentially. I had to add it because Jenkins tried running all the steps in parallel.

Jenkinsfile
pzinemanas: Because it is not only ingenialink, we'd change all the other pipelines that lock the machine in the options.

Jenkinsfile
polfeliu: Why should they run sequentially?

pzinemanas: There are issues when creating the tox environments at the same time. And besides that, the tests start a connection with the real drives, so we should have one connection type of connection at a time.

pzinemanas:
> missing doc

@polfeliu : https://novantamotion.atlassian.net/wiki/spaces/SKB/pages/553189390/Locking+machines+with+real+drives

Jenkinsfile
pzinemanas: https://novantamotion.atlassian.net/wiki/spaces/SKB/pages/553189390/Locking+machines+with+real+drives

Jenkinsfile
polfeliu: What are those issues? Matrix only runs tasks in parallel when a there are multiple nodes. If it's the same node it doesn't have any other option than to run them sequentially.

we should have one connection type of connection at a time -> Thats what the ecat_node_lock is supposed to do, isn't it?

pzinemanas: I guess. But it seems that the matrix is ignoring the lock somehow: See [this job](http://jenkins.ingenia/job/Novanta%20Motion%20-%20Ingenia%20-%20Git/job/ingeniamotion/job/test-INGM-338/2/console) for example.

GullinBustin (approved):

PR #151, opened 2024-03-06
INGM-399 Design a register poller using pd os

MartinALL91:
### Description

Create a register Poller using PDOs. NOTE: only available for FW version 2.5.0 and higher.

Fixes # INGM-399

### Type of change

- Create the PDOPoller class.
- Add the create_poller method.

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### DIFF
--- a/ingeniamotion/pdo.py
+++ b/ingeniamotion/pdo.py
@@ -18,6 +18,122 @@ if TYPE_CHECKING:
     from ingeniamotion.motion_controller import MotionController
 
 
+class PDOPoller:
+    """Poll register values using PDOs"""
+
+    DEFAULT_RPDO_REGISTER = "CL_POS_SET_POINT_VALUE"
+
+    def __init__(
+        self,
+        mc: "MotionController",
+        servo: str,
+        refresh_time: float,
+        buffer_size: int,
+    ) -> None:
+        """Constructor.
+
+        Args:
+            mc: MotionController instance
+            servo: drive alias.
+            refresh_time: PDO values refresh time.
+            buffer_size: Maximum number of register readings to store.
+
+        """
+        super().__init__()
+        self.__mc = mc
+        self.__servo = servo
+        self.__refresh_time = refresh_time
+        self.__buffer_size = buffer_size
+        self.__num_channels: int = 0
+        self.__buffer: List[List[Union[int, float]]] = [[]]
+        self.__timestamps: List[float] = []
+        self.__start_time: Optional[float] = None
+        self.__tpdo_map: TPDOMap = self.__mc.capture.pdo.create_empty_tpdo_map()
+
+    def start(self) -> None:
+        """Start the poller"""
+        self.__mc.capture.pdo.subscribe_to_receive_process_data(self._new_data_available)
+        self.__start_time = time.time()
+        self.__mc.capture.pdo.start_pdos(refresh_rate=self.__refresh_time)
+
+    def stop(self) -> None:
+        """Stop the poller"""
+        self.__mc.capture.pdo.stop_pdos()
+        self.__mc.capture.pdo.unsubscribe_to_receive_process_data(self._new_data_available)
+
+    @property
+    def data(self) -> Tuple[List[float], List[List[Union[int, float]]]]:
+        """
+        Get the poller data. After the data is retrieved, the data buffers are cleared.
+
+        Returns:
+            A list with the readings timestamps and values.
+
+        """
+        data = self.__buffer
+        time_stamps = self.__timestamps
+        self._clear_buffers()
+        return time_stamps, data
+
+    def add_channels(self, registers: List[Dict[str, Union[int, str]]]) -> None:
+        """
+        Configure the PDOs with the registers to be read.
+
+        Args:
+            registers : list of registers to add to the Poller.
+
+        """
+        self.__num_channels = len(registers)
+        self._clear_buffers()
+        for register in registers:
+            name = register.get("name", DEFAULT_SERVO)
+            if not isinstance(name, str):
+                raise ValueError(
+                    f"Wrong type for the 'name' field. Expected 'str', got: {type(name)}"
+                )
+            axis = register.get("axis", DEFAULT_AXIS)
+            if not isinstance(axis, int):
+                raise ValueError(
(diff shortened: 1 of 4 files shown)


ingeniamotion/pdo.py
```diff
@@ -18,6 +18,133 @@
...
+        self.__timestamps: List[float] = []
+        self.__start_time: Optional[float] = None
+        self.__tpdo_map: TPDOMap = self.__mc.capture.pdo.create_empty_tpdo_map()
+        self.__rpdo_map: RPDOMap = self.__mc.capture.pdo.create_empty_rpdo_map()
+        self.__fill_rpdo_map()
+
+    def start(self) -> None:
+        """Start the poller"""
+        self._clear_buffers()
+        self.__mc.capture.pdo.clear_pdo_mapping(servo=self.__servo)
```
GullinBustin, line 56: PDOs should not be cleared, it can generates conflicts with FSOE

ingeniamotion/pdo.py
```diff
@@ -18,6 +18,133 @@
...
+        self.__mc.capture.pdo.set_pdo_maps_to_slave(
+            self.__rpdo_map, self.__tpdo_map, servo=self.__servo
+        )
+        self.__mc.capture.pdo.subscribe_to_receive_process_data(self._new_data_available)
+        self.__start_time = time.time()
+        self.__mc.capture.pdo.start_pdos(refresh_rate=self.__refresh_time)
+
+    def stop(self) -> None:
```
GullinBustin, line 64: This function should remove Poller PDO map

tests/test_pdo.py
```diff
@@ -270,3 +271,31 @@ def test_start_pdos_number_of_network_exception(mocker, motion_controller):
         mc.capture.pdo.start_pdos()
     with pytest.raises(IMException):
         mc.capture.pdo.start_pdos(COMMUNICATION_TYPE.Ethercat)
+
+
+def skip_if_pdo_padding_is_not_available(mc, alias):
+    pdo_poller_fw_version = "2.5.0"
```
GullinBustin, line 277: Maybe you should modify config.json to install FW 2.5.0.

GullinBustin (changes requested):

tests/test_pdo.py
MartinALL91: Done

WoodElv (approved):

ingeniamotion/pdo.py
```diff
@@ -18,6 +18,134 @@
...
+
+    def _new_data_available(self) -> None:
+        """Add readings to the buffers.
+
+        Raises:
+            ValueError: If the poller has not been started yet.
+
+        """
+        if len(self.__timestamps) == self.__buffer_size:
+            self.__timestamps.pop(0)
```
pzinemanas, line 103: Why do you pop this element?

MartinALL91: Because the buffer is full, we need space to add the new timestamp.

ingeniamotion/pdo.py
```diff
@@ -18,6 +18,134 @@
...
+        if self.__start_time is None:
+            raise ValueError("The poller has not been started yet.")
+        time_stamp = round(time.time() - self.__start_time, 6)
+        self.__timestamps.append(time_stamp)
+        for tpdo_index, tpdo_map_item in enumerate(self.__tpdo_map.items):
+            if len(self.__buffer[tpdo_index]) == self.__buffer_size:
+                self.__buffer[tpdo_index].pop(0)
+            self.__buffer[tpdo_index].append(tpdo_map_item.value)
```
pzinemanas, line 111: Since the buffer size is fixed, we could have better performance by defining the buffer list in the constructor instead of having lists that change dynamically. Then we could have a cyclic variable to store the number of samples so that when the data is retrieved we can return `data[:number_of_samples]`. I think that reducing the time of this callback is wise so we can achieve the best performance of PDOs. What do you think?

MartinALL91: I'll use a [deque](https://docs.python.org/2/library/collections.html#collections.deque)

pzinemanas (approved):

ingeniamotion/pdo.py
```diff
@@ -18,6 +21,130 @@
...
+        self.__mc.capture.pdo.remove_rpdo_map(self.__servo, self.__rpdo_map)
+        self.__mc.capture.pdo.remove_tpdo_map(self.__servo, self.__tpdo_map)
+
+    @property
+    def data(self) -> Tuple[Deque[float], Dict[int, Deque[Union[int, float]]]]:
+        """
+        Get the poller data. After the data is retrieved, the data buffers are cleared.
+
+        Returns:
+            A list with the readings timestamps and values.
```
pzinemanas, line 78: The returns should be explained better in the docstring.

ingeniamotion/pdo.py
```diff
@@ -18,6 +21,130 @@
...
+            buffer_size: Maximum number of register readings to store.
+
+        """
+        super().__init__()
+        self.__mc = mc
+        self.__servo = servo
+        self.__refresh_time = refresh_time
+        self.__buffer_size = buffer_size
+        self.__num_channels: int = 0
+        self.__buffer: Dict[int, Deque[Union[int, float]]] = defaultdict(partial(deque, maxlen=self.__buffer_size))  # type: ignore
```
GullinBustin, line 49: Why do you use `defaultdict`? I think this should be a list, as in the Poller, and it should be initialized in the start function.

GullinBustin (changes requested):

ingeniamotion/pdo.py
MartinALL91: I don't think it makes a difference if it's a list or a dictionary. Both pollers don't need the same internals, as long as they work similarly.

GullinBustin: I think we should avoid the use `# type: ignore` if it is not for a good reason.

MartinALL91: Done

ingeniamotion/pdo.py
MartinALL91: Done

GullinBustin (approved):

PR #157, opened 2024-03-28
INGM-407: make new example about can bootloader

WoodElv:
### Description

A new script is added into the examples folder for loading a firmware using CAN bootloader.
Also, unit tests are added for this script.

### Type of change

- [x] Script for loading a firmware using CAN bootloader.

### Tests
- [x] Unit test if the script successes.
- [x] Unit test if the script failed.

### Documentation

### Code formatting


examples/load_fw_canopen.py
```diff
@@ -0,0 +1,79 @@
...
+def status_log(current_status: str) -> None:
+    print(f"Load firmware status: {current_status}")
+
+
+def progress_log(current_progress: int) -> None:
+    print(f"Load firmware progress: {current_progress}%")
+
+
+def load_firmware_canopen(can_drive: Dict[str, Any]) -> None:
+    # Create MotionController instance
```
pzinemanas, line 18: This comment does not provide much information, it could be removed.

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,79 @@
...
+    mc = MotionController()
+    # Find available nodes
+    node_id_list = mc.communication.scan_servos_canopen(
+        can_drive["device"],
+        baudrate=can_drive["baudrate"],
+        channel=can_drive["channel"],
+    )
+    if can_drive["node_id"] not in node_id_list:
+        print("No nodes are connected.")
```
pzinemanas, line 27: This message is not correct. e.g.:
`print(f"Node {can_drive["node_id"]} is not detected.")`

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,79 @@
...
+        return
+    else:
+        print(f"Found nodes: {node_id_list}")
+
+    # Connect drive
+    print("Starts to established a communication.")
+    alias_can_servo = "custom_alias"
```
pzinemanas, line 34: Why do you use a custom alias? Could not be the default one for simplicity?

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,79 @@
...
+        channel=can_drive["channel"],
+        alias=alias_can_servo,
+    )
+    if alias_can_servo in mc.servos:
+        print("Drive is connected.")
+    else:
+        print("Drive is not connected.")
+        return
+
+    # Load firmware
```
pzinemanas, line 49: Remove comment.

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,79 @@
...
+    try:
+        mc.communication.load_firmware_canopen(
+            can_drive["fw_path"],
+            servo=alias_can_servo,
+            status_callback=status_log,
+            progress_callback=progress_log,
+        )
+        print("Firmware is uploaded successfully.")
+    except ILFirmwareLoadError as e:
+        print(f"CAN boot error: {e}")
```
pzinemanas, line 60: Firmware loading failed.

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,79 @@
...
+    finally:
+        try:
+            mc.communication.disconnect(alias_can_servo)
+            print("Drive is disconnected.")
+        except Exception as e:
+            print(f"Error when disconnection from drive: {e}")
```
pzinemanas, line 66: Error during drive disconnection?

examples/load_fw_canopen.py
WoodElv: It's easier without a custom alias, but we don't have a lot of examples using a custom alias. I think it's good having some to teach how to use them. What do you think?

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,76 @@
...
+            print(f"Error during drive disconnection: {e}")
+
+
+if __name__ == "__main__":
+    # Remember to replace all parameters here
+    can_drive = {
+        "device": CAN_DEVICE.KVASER,
+        "channel": 0,
+        "baudrate": CAN_BAUDRATE.Baudrate_1M,
+        "node_id": 20,
```
MartinALL91, line 72: I would set the node_id to the default value of our drives (32).

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,76 @@
...
+        baudrate=can_drive["baudrate"],
+        channel=can_drive["channel"],
+    )
+    if can_drive["node_id"] not in node_id_list:
+        print(f"Node {can_drive['node_id']} is not detected.")
+        return
+    else:
+        print(f"Found nodes: {node_id_list}")
+
+    print("Starts to established a communication.")
```
MartinALL91, line 31: establish

pzinemanas (approved):

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,73 @@
...
+
+    print("Starts to establish a communication.")
+    mc.communication.connect_servo_canopen(
+        can_drive["device"],
+        can_drive["dictionary_path"],
+        can_drive["node_id"],
+        baudrate=can_drive["baudrate"],
+        channel=can_drive["channel"],
+    )
+    if "default" in mc.servos:
```
mjedieckmann, line 39: should probably be try / catch instead

mjedieckmann (approved):

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,73 @@
...
+    # Remember to replace all parameters here
+    can_drive = {
+        "device": CAN_DEVICE.KVASER,
+        "channel": 0,
+        "baudrate": CAN_BAUDRATE.Baudrate_1M,
+        "node_id": 32,
+        "dictionary_path": "parent_directory/full_dictionary_path.xdf",
+        "fw_path": "parent_directory/full_firmware_file_path.lfu",
+    }
+    load_firmware_canopen(can_drive)
```
GullinBustin, line 73: Why do you use a dict instead use different arguments?

GullinBustin (changes requested):

examples/load_fw_canopen.py
WoodElv: Because there are a lot of parameters, and I think it's cleaner passing just one input parameter instead of 6 parameters for a script.

GullinBustin: In my opinion, it is weird. A user can't know which parameters have the function seeing the declaration, an example of use is necessary to know all the parameters or see the full implementation, which is worse. A lot of customers will use these examples as templates to do their code, we must follow good practices.

tests/test_examples.py
```diff
@@ -212,3 +214,89 @@ def enable_brake(*args, **kwargs):
     mocker.patch.object(MotionController, "configuration", MockConfiguration)
     result = script_runner.run(script_path, override, dictionary, f"-ip={ip_address}")
     assert result.returncode == 0
+
+
+@pytest.mark.virtual
+def test_can_bootloader_example_success(mocker, capsys):
+    device = "my_can_device"
+    channel = 111
+    baudrate = 4444
```
MartinALL91, line 223: These values are weird and in some cases make no sense. I would use normal ones.

examples/load_fw_canopen.py
WoodElv: I'm going to remove this conditional statement because the no connection state can't be reached.

examples/load_fw_canopen.py
```diff
@@ -0,0 +1,68 @@
...
+            progress_callback=progress_log,
+        )
+        print("Firmware is uploaded successfully.")
+    except ILFirmwareLoadError as e:
+        print(f"Firmware loading failed: {e}")
+    finally:
+        try:
+            mc.communication.disconnect()
+            print("Drive is disconnected.")
+        except Exception as e:
```
GullinBustin, line 56: Why did you add a try-except in the disconnect function? and why is it a generic exception?

GullinBustin (changes requested):

examples/load_fw_canopen.py
WoodElv: I put it because if there was an unhandled exception. But I'm going to take it off because in that cases an exception should to be raised.

GullinBustin (approved):

PR #158, opened 2024-04-02
Ingm 404 make new example about ether cat co e connection

WoodElv:
### Description

A new script is added into the examples folder for a drive connection using EtherCAT - CoE.
Also, unit tests are added for this script.

### Type of change

- [x] Script for a drive connection using EtherCAT - CoE.

### Tests
- [x] Unit test if the CoE connection successes.
- [x] Unit test if the CoE connection failed.
- [x] Unit test if there is a ConnectionError exception.

### Documentation

### Code formatting


examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,63 @@
...
+
+    try:
+        slave_id_list = mc.communication.scan_servos_ethercat(interface_selected)
+    except ConnectionError as e:
+        print(e)
+        return
+
+    if not slave_id_list:
+        print(
+            f"Any slave is detected using the interface: {interface_list_human_format[ecat_coe_conf['interface_index']]}."
```
MartinALL91, line 30: No slave detected on interface

examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,63 @@
...
+        print(f"Found slaves: {slave_id_list}")
+
+    try:
+        mc.communication.connect_servo_ethercat(
+            interface_selected, ecat_coe_conf["slave_id"], ecat_coe_conf["dictionary"]
+        )
+    except FileNotFoundError as e:
+        print(e)
+        return
+    if "default" not in mc.servos:
```
MartinALL91, line 43: Something has had to go wrong to reach this point. I would expand the try/except block.

examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,63 @@
...
+        print("Drive is not connected.")
+        return
+    else:
+        print("Drive is connected.")
+
+    mc.communication.disconnect()
+    if "default" in mc.servos:
```
MartinALL91, line 50: In what case could this fail?

examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,63 @@
...
+        print("Drive disconnection failed.")
+    else:
+        print("The drive has been disconnected.")
+
+
+if __name__ == "__main__":
+    # Replace the ecat_coe_conf parameters
+    ecat_coe_conf = {
+        "interface_index": 3,
```
mjedieckmann, line 59: Can we hardcode the index like that? Aren't the available interfaces and their indices dependent on the machine you run the code on?

mjedieckmann (approved):

examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,63 @@
...
+    print(f"- Index interface: {ecat_coe_conf['interface_index']}")
+    print(f"- Real name: {interface_selected}")
+    print(
+        f"- Human-readable format name: {interface_list_human_format[ecat_coe_conf['interface_index']]}"
+    )
+
+    try:
+        slave_id_list = mc.communication.scan_servos_ethercat(interface_selected)
+    except ConnectionError as e:
+        print(e)
```
GullinBustin, line 25: Why do you use a try-except instead of do nothing and allow raise an exception if something is wrong? It is not a good practice.

GullinBustin (changes requested):

examples/connect_ecat_coe.py
GullinBustin: Did you think about use logger or ingenialogger instead prints?

WoodElv: Yes, but there are another loggers that come from the library, and for an example script I think they make noise. I guess that's the reason that there is any logger in any other example that is already made.

examples/connect_ecat_coe.py
WoodElv: Done

examples/connect_ecat_coe.py
WoodElv: Done

examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,42 @@
+from ingeniamotion import MotionController
+
+
+def connect_ethercat_coe(interface_index: int, slave_id: int, dictionary_path: str) -> None:
+    mc = MotionController()
+
+    interface_list_human_format = mc.communication.get_interface_name_list()
+    print("List of interfaces - Human-readable format:")
```
pzinemanas, line 8: I'd remove the "Human-readable format"

examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,42 @@
...
+    for index, interface in enumerate(interface_list_human_format):
+        print(f"{index}: {interface}")
+
+    interface_selected = mc.communication.get_ifname_by_index(interface_index)
+    print("Interface selected:")
+    print(f"- Index interface: {interface_index}")
+    print(f"- Real name: {interface_selected}")
```
pzinemanas, line 15: Interface identifier?

examples/connect_ecat_coe.py
GullinBustin: It is an example, not a script. Add prints or logs could be good, but sometimes adding comments in the code is better than adding prints. No changes are required, but I think it is good to remember for future examples.

GullinBustin (approved):

examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,42 @@
+from ingeniamotion import MotionController
+
+
+def connect_ethercat_coe(interface_index: int, slave_id: int, dictionary_path: str) -> None:
+    mc = MotionController()
+
+    interface_list_human_format = mc.communication.get_interface_name_list()
```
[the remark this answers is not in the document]
WoodElv: Done

examples/connect_ecat_coe.py
WoodElv: Done

examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,42 @@
...
+    print("List of interfaces - Human-readable format:")
+    for index, interface in enumerate(interface_list_human_format):
+        print(f"{index}: {interface}")
+
+    interface_selected = mc.communication.get_ifname_by_index(interface_index)
+    print("Interface selected:")
+    print(f"- Index interface: {interface_index}")
+    print(f"- Real name: {interface_selected}")
+    print(f"- Human-readable format name: {interface_list_human_format[interface_index]}")
```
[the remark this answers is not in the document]
WoodElv: Done

MartinALL91 (approved):

pzinemanas (approved):

examples/connect_ecat_coe.py
```diff
@@ -0,0 +1,42 @@
...
+        print(f"No slave detected on interface: {interface_list[interface_index]}")
+        return
+    else:
+        print(f"Found slaves: {slave_id_list}")
+
+    try:
+        mc.communication.connect_servo_ethercat(interface_selected, slave_id, dictionary_path)
+    except FileNotFoundError as e:
+        print(e)
+        return
```
GullinBustin, line 30: Why did you add a try-except? I think it's better to raise the exception. The behavior is very similar, but now seems that the file not found is an expected behavior.

GullinBustin (changes requested):

GullinBustin (approved):

PR #167, opened 2024-04-12
INGM-408: make new example about save and load config

WoodElv:
### Description

The example load and save a configuration, and checks how it affects to Max. Velocity register value.

Fixes # INGM-408

### Type of change

- [x] Basic Load/Save configuration example
- [x] Load/Save configuration with register changes example

### Tests
- [x] Run basic Load/Save configuration example.
- [x] Run Load/Save configuration with register changes example successfully.
- [x] Run Load/Save configuration with register changes example failed.

### Documentation

### Code formatting

### DIFF
--- a/examples/load_save_config_register_changes.py
+++ b/examples/load_save_config_register_changes.py
@@ -6,9 +6,7 @@ def main() -> None:
     # Modify these parameters to connect a drive
     interface_index = 3
     slave_id = 1
-    dictionary_path = (
-        "parent_directory/dictionary_file.xdf"
-    )
+    dictionary_path = "parent_directory/dictionary_file.xdf"
     mc.communication.connect_servo_ethercat_interface_index(
         interface_index, slave_id, dictionary_path
     )
--- a/examples/load_save_config_register_changes.py
+++ b/examples/load_save_config_register_changes.py
@@ -1,0 +1,48 @@
+from ingeniamotion import MotionController
+
+
+def main() -> None:
+    mc = MotionController()
+    # Modify these parameters to connect a drive
+    interface_index = 3
+    slave_id = 1
+    dictionary_path = (
+        "parent_directory/dictionary_file.xdf"
+    )
+    mc.communication.connect_servo_ethercat_interface_index(
+        interface_index, slave_id, dictionary_path
+    )
+    # Save the initial configuration of your drive in a file
+    initial_config_path = "initial_configuration.xcf"
+    mc.configuration.save_configuration(initial_config_path)
+    print("The initial configuration is saved.")
+
+    # Get the initial max. velocity and set it with a new value
+    new_max_velocity = 20.0
+    initial_max_velocity = mc.configuration.get_max_velocity()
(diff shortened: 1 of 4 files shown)


examples/load_save_configuration.py
```diff
@@ -0,0 +1,48 @@
+from ingeniamotion import MotionController
+
+
+def main() -> None:
```
GullinBustin, line 4: I think this example could be confusing for someone who wants to load and save a config. It seems more like a test than an example. What do you think?

WoodElv: I agree, a basic example is necessary, but I'm going to keep the current example as well. The reason is posted in the description issue.

WoodElv: Done!

GullinBustin (approved):

examples/load_save_config_register_changes.py
```diff
@@ -0,0 +1,46 @@
...
+    mc.configuration.set_max_velocity(new_max_velocity)
+
+    # Save the configuration with changes in a file
+    modified_config_path = "configuration_example.xcf"
+    mc.configuration.save_configuration(modified_config_path)
+    print("The configuration file is saved with the modification.")
+
+    # Load the initial configuration and check the max. velocity register has its initial value
+    mc.configuration.load_configuration(initial_config_path)
+    if mc.configuration.get_max_velocity() == initial_max_velocity:
```
mjedieckmann, line 34: Instead of doing this in an if-statement, I would just print the result and what we expect in any case, such that the user can easily see if what they are trying to do worked.
print(f"Max. velocity register should be set to its initial value ({initial_max_velodity}). Current value: {mc.configuration.get_max_velocity()}")

examples/load_save_config_register_changes.py
```diff
@@ -0,0 +1,46 @@
...
+        print("Max. velocity register has the initial value.")
+
+    # Load the modified configuration and check the max. velocity value is changed again
+    mc.configuration.load_configuration(modified_config_path)
+    if mc.configuration.get_max_velocity() == new_max_velocity:
```
mjedieckmann, line 39: See comment above and change to something like:
print(f"Max. velocity register should now be set to the new value ({new_max_velodity}). Current value: {mc.configuration.get_max_velocity()}")

tests/test_examples.py
```diff
@@ -301,3 +307,58 @@ def disconnect(*args, **kwargs):
...
+    mocker.patch.object(Configuration, "set_max_velocity")
+
+    main_load_save_config_register_changes()
+
+    captured_outputs = capsys.readouterr()
+    all_outputs = captured_outputs.out.split("\n")
+
+    assert all_outputs[0] == "The initial configuration is saved."
+    assert all_outputs[1] == "The configuration file is saved with the modification."
+    assert all_outputs[2] == "Max. velocity register has the initial value."
```
mjedieckmann, line 345: these two lines will then have to be updated

mjedieckmann (changes requested):

PR #177, opened 2024-05-10
INGM-367 Use a queue to manage stoppable sleeps in the wizard tests

pzinemanas:
### Description

In this PR, a queue is used to manage the stoppable sleeps in the wizard test so the main Python thread is not blocked.

Fixes # INGM-367

### Type of change

- [x] Use a queue to manage stoppable sleeps in the wizard tests

### Tests
- [x] Test that the brake tune tests in ML3 does not blocks the GUI.
- [x] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/wizard_tests/stoppable.py
+++ b/ingeniamotion/wizard_tests/stoppable.py
@@ -12,7 +12,6 @@ T = typing.TypeVar("T")
 
 
 class Stoppable:
-
     stop_queue = Queue(1)
 
     @staticmethod
--- a/ingeniamotion/wizard_tests/stoppable.py
+++ b/ingeniamotion/wizard_tests/stoppable.py
@@ -1,7 +1,7 @@
 import typing
-from time import time
 from functools import wraps
-from typing import TypeVar, Callable, Any
+from queue import Empty, Queue
+from typing import Callable
 
 
 class StopException(Exception):
@@ -13,7 +13,7 @@ T = typing.TypeVar("T")
 
 class Stoppable:
 
-    is_stopped = False
+    stop_queue = Queue(1)
 
     @staticmethod
     def stoppable(fun: Callable[..., T]) -> Callable[..., T]:
@@ -25,16 +25,21 @@ class Stoppable:
         return wrapper
 
     def reset_stop(self) -> None:
-        self.is_stopped = False
+        if self.stop_queue.full():
+            self.stop_queue.get(block=False)
 
     def stop(self) -> None:
-        self.is_stopped = True
+        if not self.stop_queue.full():
+            self.stop_queue.put(StopException, block=False)
 
     def check_stop(self) -> None:
-        if self.is_stopped:
-            raise StopException
+        if self.stop_queue.full():
+            raise self.stop_queue.get(block=False)
 
     def stoppable_sleep(self, timeout: float) -> None:
-        init_time = time()
-        while init_time + timeout > time():
-            self.check_stop()
+        try:
+            stop_exception = self.stop_queue.get(block=True, timeout=timeout)
+        except Empty:
+            pass
+        else:
+            raise stop_exception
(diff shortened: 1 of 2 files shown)


MartinALL91 (approved):

ingeniamotion/wizard_tests/stoppable.py
```diff
@@ -24,16 +24,21 @@ def wrapper(self, *args, **kwargs):  # type: ignore
...
     def stop(self) -> None:
-        self.is_stopped = True
+        if not self.stop_queue.full():
+            self.stop_queue.put(StopException(), block=False)
 
     def check_stop(self) -> None:
-        if self.is_stopped:
-            raise StopException
+        if self.stop_queue.full():
+            raise self.stop_queue.get(block=False)
```
polfeliu, line 36: First checking the list and then getting the item could create a race condition where get(block=False) would return None and raise None.
Simple solution:

```
def check_stop(self) -> None:
  ex = self.stop_queue.get(block=False)
  
  if ex is not None:
      raise ex
```
Also, mypy is not complaining that you could raise None? isn't strictness set to True for this project??

ingeniamotion/wizard_tests/stoppable.py
```diff
@@ -24,16 +24,21 @@ def wrapper(self, *args, **kwargs):  # type: ignore
         return wrapper
 
     def reset_stop(self) -> None:
-        self.is_stopped = False
+        if self.stop_queue.full():
+            self.stop_queue.get(block=False)
```
polfeliu, line 28: Same race condition, just deleted the .full() condition

ingeniamotion/wizard_tests/stoppable.py
```diff
@@ -24,16 +24,21 @@ def wrapper(self, *args, **kwargs):  # type: ignore
...
 
     def stop(self) -> None:
-        self.is_stopped = True
+        if not self.stop_queue.full():
+            self.stop_queue.put(StopException(), block=False)
```
polfeliu, line 32: Same. Just put(exception, block=False)

polfeliu:
Are the specific unit tests for this method? Or are implicitly validated by others?

pzinemanas:
> Are the specific unit tests for this method? Or are implicitly validated by others?

@polfeliu, yes there are tests like `test_feedback_stop`

ingeniamotion/wizard_tests/stoppable.py
pzinemanas: Mypy is not complaining because the get method raises an Empty exception if the queue is empy instead of returning None. That's why I'm checking if the queue is full before calling the get method. But I see your point, I'll fix it.

https://docs.python.org/3/library/queue.html#queue.Queue.get

pzinemanas: @polfeliu Done [fe1299d](https://github.com/ingeniamc/ingeniamotion/pull/177/commits/fe1299dffcffa1751c4b9862a4b49d88d2512d25)

ingeniamotion/wizard_tests/stoppable.py
pzinemanas: @polfeliu Done [fe1299d](https://github.com/ingeniamc/ingeniamotion/pull/177/commits/fe1299dffcffa1751c4b9862a4b49d88d2512d25)

ingeniamotion/wizard_tests/stoppable.py
pzinemanas: @polfeliu Done [fe1299d](https://github.com/ingeniamc/ingeniamotion/pull/177/commits/fe1299dffcffa1751c4b9862a4b49d88d2512d25)

polfeliu (approved):

GullinBustin (approved):

PR #180, opened 2024-05-22
INGM-448 Import ingenialink enums into the enum module

GullinBustin:
### Description

Add ingenialink enums to ingeniamotion enums module

Fixes # INGM-448

### Type of change

- [x] Add ingenialink enums to enums module (description)

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/enums/__init__.py
+++ b/ingeniamotion/enums/__init__.py
@@ -1,5 +1,8 @@
 from enum import Enum, IntEnum
 
+from ingenialink.canopen.network import CAN_BAUDRATE, CAN_DEVICE
+from ingenialink.enums.register import REG_ACCESS, REG_DTYPE
+
 
 class OperationMode(IntEnum):
     """Operation Mode Enum"""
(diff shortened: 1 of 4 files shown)


MartinALL91 (approved):

ingeniamotion/enums/__init__.py
```diff
@@ -209,8 +212,3 @@ class FilterNumber(IntEnum):
 
     FILTER1 = 1
     FILTER2 = 2
-
-
-enums = list(globals().keys())
-enums.remove("IntEnum")
-__all__ = enums
```
polfeliu, line 216: Beware that now the users that were doing
`from ingeniamotion.enums import *`

I don't see the problem of listing the enums that are public on the top of the file.
According to the PEP8 all these enums should be considered internal...
https://peps.python.org/pep-0008/#public-and-internal-interfaces

GullinBustin: Now we can only add a new enum and it will work fine. If we use __all__, we must add the enum to __all__, this is an unnecessary additional step.

polfeliu (approved):

PR #181, opened 2024-05-27
INGM-433 Example script to check servo configurations

mjedieckmann:
### Description

Fixes # (https://novantamotion.atlassian.net/browse/INGM-433)

### Type of change

- [x] Add check_configuration function to ingeniamotion
- [x] Provide example file on how to use it to check the configuration of a given list of servos

### Documentation

### Code formatting

- [x] Use ruff to format the code

### DIFF
--- a/ingeniamotion/configuration.py
+++ b/ingeniamotion/configuration.py
@@ -161,6 +161,35 @@ class Configuration(Homing, Feedbacks, metaclass=MCMetaClass):
         """
         self.disable_brake_override(servo, axis)
 
+    def check_configuration(
+        self, config_path: str, axis: Optional[int] = None, servo: str = DEFAULT_SERVO
+    ) -> None:
+        """Check if the drive is configured in the same way as the given configuration file.
+        Compares the value of each register in the given file with the corresponding value in the
+        drive.
+
+        Args:
+            config_path : config file path to check.
+            axis : target axis to load configuration.
+                If ``None`` function loads all axis. ``None`` by default.
+            servo : servo alias to reference it. ``default`` by default.
+
+        Raises:
+            FileNotFoundError: If configuration file does not exist.
+            ValueError: If a configuration file from a subnode different from 0
+                is attempted to be loaded to subnode 0.
+            ValueError: If an invalid subnode is provided.
+            ILConfigurationError: If the configuration file differs from the drive state.
+
+        """
+        if not path.isfile(config_path):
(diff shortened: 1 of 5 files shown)


examples/check_configurations_example.py
```diff
@@ -0,0 +1,65 @@
...
+    print("Starting script to check configurations..")
+    print("Checking for each servo if their current configuration matches the given file.")
+    # The folder where you placed the dictionary and configuration - files for the servos
+    # you defined in SERVOS_TO_TEST.
+    resources_folder = "examples/resources/"
+    mc = MotionController()
+
+    for servo_to_test in SERVOS_TO_TEST:
+        ip = servo_to_test["ip"]
+        config = resources_folder + servo_to_test["config"]
```
pzinemanas, line 39: Use `os.path.join` or similar to join paths,

pzinemanas (changes requested):
Please add a test in test_examples.py.

ingeniamotion/configuration.py
```diff
@@ -161,6 +161,35 @@ def default_brake(self, servo: str = DEFAULT_SERVO, axis: int = DEFAULT_AXIS) ->
...
+        Raises:
+            FileNotFoundError: If configuration file does not exist.
+            ValueError: If a configuration file from a subnode different from 0
+                is attempted to be loaded to subnode 0.
+            ValueError: If an invalid subnode is provided.
+            ILConfigurationError: If the configuration file differs from the drive state.
+
+        """
+        if not path.isfile(config_path):
+            raise FileNotFoundError("{} file does not exist!".format(config_path))
```
MartinALL91, line 186: Use an f-string

ingeniamotion/configuration.py
```diff
@@ -161,6 +161,35 @@ def default_brake(self, servo: str = DEFAULT_SERVO, axis: int = DEFAULT_AXIS) ->
         """
         self.disable_brake_override(servo, axis)
 
+    def check_configuration(
+        self, config_path: str, axis: Optional[int] = None, servo: str = DEFAULT_SERVO
+    ) -> None:
```
MartinALL91, line 166: I think It would be best if the method return a boolean.

mjedieckmann: The issue specified that the ingenialink function should throw an exception when the check fails:
image
I could change this function to catch the error and return a boolean, but I'm not sure if this is what we need.

GullinBustin (approved):

PR #183, opened 2024-05-29
INGM-453 Update the FSoE module

MartinALL91:
### Description

Update the FSoE module with the changes in the API.

Fixes # INGM-453

### Type of change

- Update the stop method.
- Delete the FSoE master handler upon disconnection.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/communication.py
+++ b/ingeniamotion/communication.py
@@ -827,6 +827,7 @@ class Communication(metaclass=MCMetaClass):
         del self.mc.servos[servo]
         net_name = self.mc.servo_net.pop(servo)
         servo_count = list(self.mc.servo_net.values()).count(net_name)
+        self.mc.fsoe._delete_master_handler(servo)
         if servo_count == 0:
             del self.mc.net[net_name]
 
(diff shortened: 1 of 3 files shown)


pzinemanas (approved):

GullinBustin (approved):

tox.ini
```diff
@@ -8,6 +8,7 @@ env_list = coverage, build, firmware, docs, format, type, py{39,310,311,312}, vi
 # It should be empty in production.
 [develop]
     ingenialink = git+https://github.com/ingeniamc/ingenialink-python@develop
+    # TODO: Update the commit ref once the open PRs are merged
```
polfeliu, line 11: they are now

ingeniamotion/communication.py
```diff
@@ -827,6 +828,9 @@ def disconnect(self, servo: str = DEFAULT_SERVO) -> None:
         del self.mc.servos[servo]
         net_name = self.mc.servo_net.pop(servo)
         servo_count = list(self.mc.servo_net.values()).count(net_name)
+        # TODO: Remove once INGK-912 is resolved
+        with contextlib.suppress(NotImplementedError):
```
polfeliu, line 832: Why is this dependant on the release?
Create a task to do this or add it to the issue, otherwise we might forget

MartinALL91: Because right now the fsoe_master library is only being installed on one machine (the one that is able to compile it).

PR #188, opened 2024-06-04
INGM-457 Add method to check the STO state in safe drives

MartinALL91:
### Description

Add a method to check the STO state in safe drives

Fixes # INGM-457

### Type of change

- Add a method to check the STO state in safe drives.
- Update the STO example script.

### Tests
- Test the example script with a safe drive.
- Check that it works properly.

### Documentation

### Code formatting

### Others

### DIFF
--- a/examples/safety_torque_off.py
+++ b/examples/safety_torque_off.py
@@ -1,5 +1,4 @@
 import contextlib
-import time
 
 from ingeniamotion import MotionController
 from ingeniamotion.enums import OperationMode
@@ -22,7 +21,7 @@ def main(interface_ip, slave_id, dict_path):
     mc.fsoe.sto_deactivate()
     # Wait for the STO to be deactivated
     while mc.fsoe.check_sto_active():
-        time.sleep(0.1)
+        pass
     # Enable the motor
     mc.motion.motor_enable()
     # Wait for the motor to reach a certain velocity (10 rev/s)
--- a/examples/safety_torque_off.py
+++ b/examples/safety_torque_off.py
@@ -1,4 +1,5 @@
 import contextlib
+import time
 
 from ingeniamotion import MotionController
 from ingeniamotion.enums import OperationMode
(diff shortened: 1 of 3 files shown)


ingeniamotion/fsoe.py
```diff
@@ -92,6 +100,21 @@ def sto_activate(self) -> None:
...
+    def is_sto_active(self) -> bool:
+        """Check the STO state.
+
+        Returns:
+            True if the STO is active. False otherwise.
+
+        """
+        if self.__master_handler.state != StateData:
+            return True
+        fsoe_command = int.from_bytes(self.safety_slave_pdu_map.items[0].raw_data_bytes, "little")
```
polfeliu, line 112: This should be a get() from a dictionary item, but since its Input and Output at the same time is not implemented.

Create issue or TODO for when this is done:
https://novantamotion.atlassian.net/browse/INGK-919

ingeniamotion/fsoe.py
```diff
@@ -92,6 +100,21 @@ def sto_activate(self) -> None:
...
+        if fsoe_command != self.PROCESS_DATA_COMMAND:
+            return True
+        sto_command = self.safety_slave_pdu_map.items[1].raw_data_bits.any()
```
polfeliu, line 115: This should be a get() from a dictionary item, but since its Input and Output at the same time is not implemented.

Create issue or TODO for when this is done:
https://novantamotion.atlassian.net/browse/INGK-919

ingeniamotion/fsoe.py
polfeliu, line 112: This should be much simpler.
https://novantamotion.atlassian.net/browse/INGK-920

MartinALL91: Done

ingeniamotion/fsoe.py
MartinALL91: Done

examples/safety_torque_off.py
```diff
@@ -17,8 +17,14 @@ def main(interface_ip, slave_id, dict_path):
     # Create and start the FSoE master handler
     mc.fsoe.create_fsoe_master_handler()
     mc.fsoe.start_master(start_pdos=True)
-    # Deactivate the STO and enable the motor
+    # Wait for the master to reach the Data state
+    mc.fsoe.wait_for_state_data()
```
polfeliu, line 21: Don't you want to set a timeout on the example?

MartinALL91: I don't think it's necessary.

polfeliu (approved):

GullinBustin (approved):

PR #189, opened 2024-06-04
INGM-458 Get the STO command value from the FSoE master dictionary

MartinALL91:
### Description

Get the STO command value from the FSoE master dictionary

Fixes # (issue)

### Type of change

- Get the STO command value from the FSoE master dictionary

### Tests
- Run the STO example script.
- Check that it works properly.

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/fsoe.py
+++ b/ingeniamotion/fsoe.py
@@ -8,6 +8,7 @@ from fsoe_master.fsoe_master import (
     DictionaryItem,
     MasterHandler,
     StateData,
+    DictionaryItemInputOutput,
 )
 from ingenialink.enums.register import REG_ACCESS, REG_DTYPE
 from ingenialink.ethercat.register import EthercatRegister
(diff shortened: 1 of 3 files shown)


polfeliu (approved):

ingeniamotion/fsoe.py
```diff
@@ -115,8 +117,9 @@ def is_sto_active(self) -> bool:
         fsoe_command = int.from_bytes(self.safety_slave_pdu_map.items[0].raw_data_bytes, "little")
```
polfeliu, line 117: You should not need to check the state of the the state machine or cehck if is in failsafe...
If it's not in process data safe inputs on the dictionary must be set to failsafe
https://novantamotion.atlassian.net/browse/INGK-922

polfeliu: With this PR: https://bitbucket.org/novantamotion/fsoe_master_python/pull-requests/39/overview
I think this should be enough
````
def is_sto_active(self) -> bool:
        """Check the STO state.

        Returns:
            True if the STO is active. False otherwise.

        """
        sto_command = self.__master_handler.dictionary.get(self.STO_COMMAND_UID)
        if not isinstance(sto_command, bool):
            raise ValueError(f"Wrong value type. Expected type bool, got {type(sto_command)}")
        return sto_command
````

GullinBustin (approved):

PR #190, opened 2024-06-05
INGM-460 Add Lock mechanism in the PDO poller

pzinemanas:
### Description

In this PR, a lock mechanism is added to the PDO poller so the samples can not be modified during the data is being read.

Fixes # INGM-460

### Type of change

- [x] Add Lock mechanism in the PDO poller

### Tests
- [x] Test it with ML3.
- [x] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Set the boot_in_app flag according to the file extension.
 - FSoE module.
 - STO example for Safe servo drives.
+- Lock mechanism in the PDO Poller.
 
 ### Fix
 - Bug retrieving interface adapter name in Linux.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Set the boot_in_app flag according to the file extension.
 - FSoE module.
 - STO example for Safe servo drives.
-- Lock mechanism in the PDO Poller.
+- Use only one queue in the PDO Poller to store both the timestamps and the register values.
 
 ### Fix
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Set the boot_in_app flag according to the file extension.
 - FSoE module.
 - STO example for Safe servo drives.
+- Lock mechanism in the PDO Poller.
```
MartinALL91, line 8: This should be updated.

ingeniamotion/pdo.py
```diff
@@ -98,9 +88,14 @@ def data(self) -> Tuple[Deque[float], List[Deque[Union[int, float]]]]:
             the readings values.
 
         """
-        data = deepcopy(self.__buffer)
-        time_stamps = deepcopy(self.__timestamps)
-        self._clear_buffers()
+        time_stamps = []
```
polfeliu, line 91: Will motionlab have to reprocess this data?
If you think the data would be better presented to it in the buffer form please create an issue.
We could potentially leave this for retrocompatibility but improve the performance of the scope

polfeliu (approved):

ingeniamotion/pdo.py
pzinemanas: I kept compatibility with the other Poller and Monitoring, so the scope receives the data in similar formats. But we could consider improving it. We need to improve the other poller as well.

PR #199, opened 2024-06-17
INGM-464 Delete network from dictionary if connection is not successful

MartinALL91:
### Description

If a connection to a drive fails, the Network instance remains in the network dictionary. Remove it from the network dictionary if no other drives are connected to it.

Fixes # INGM-464

### Type of change
- Delete the network from the dictionary if the connection is not successful.

### Tests
- Follow the steps described in the issue.
- Check that no error occurs.

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/communication.py
+++ b/ingeniamotion/communication.py
@@ -253,14 +253,19 @@ class Communication(metaclass=MCMetaClass):
         if ifname not in self.mc.net:
             self.mc.net[ifname] = EoENetwork(ifname)
         net = self.mc.net[ifname]
-        servo = net.connect_to_slave(
-            slave,
-            ip,
-            dict_path,
-            port,
-            servo_status_listener=servo_status_listener,
-            net_status_listener=net_status_listener,
-        )
+        try:
+            servo = net.connect_to_slave(
+                slave,
+                ip,
+                dict_path,
+                port,
+                servo_status_listener=servo_status_listener,
+                net_status_listener=net_status_listener,
+            )
+        except ILError as e:
+            if len(net.servos) == 0:
+                del self.mc.net[ifname]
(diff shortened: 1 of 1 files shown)


polfeliu:
I don't understand how this relates to the original issue, I may be missing on some context

pzinemanas (approved):

MartinALL91:
> I don't understand how this relates to the original issue, I may be missing on some context

When trying to connect to an EtherCAT device (via CoE or EoE) the network instance was stored even if the connection failed.  So when trying to connect to the same device but with a different protocol the previous network instance was being used. This caused an exception due to the difference in the arguments in the connect_to_slave method.

polfeliu:
> > I don't understand how this relates to the original issue, I may be missing on some context
> 
> When trying to connect to an EtherCAT device (via CoE or EoE) the network instance was stored even if the connection failed. So when trying to connect to the same device but with a different protocol the previous network instance was being used. This caused an exception due to the difference in the arguments in the connect_to_slave method.

What if you connect to one CoE device succesfully, and then fail to connect to a second CoE device? Will the network be destroyed?

MartinALL91:
> > > I don't understand how this relates to the original issue, I may be missing on some context
> > 
> > 
> > When trying to connect to an EtherCAT device (via CoE or EoE) the network instance was stored even if the connection failed. So when trying to connect to the same device but with a different protocol the previous network instance was being used. This caused an exception due to the difference in the arguments in the connect_to_slave method.
> 
> What if you connect to one CoE device succesfully, and then fail to connect to a second CoE device? Will the network be destroyed?

It will not be destroyed. It is only removed if the network has no servos connected to it.

polfeliu (approved):

PR #214, opened 2024-07-17
INGM-474 Add method to get a servo's FSoE master handler state

MartinALL91:
### Description

Add a method to get a servo's FSoE master handler state. 

Fixes # INGM-474

### Type of change

- Add a method to get a servo's FSoE master handler state.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/fsoe.py
+++ b/ingeniamotion/fsoe.py
@@ -8,6 +8,7 @@ from fsoe_master.fsoe_master import (
     DictionaryItem,
     DictionaryItemInputOutput,
     MasterHandler,
+    State,
     StateData,
 )
 from ingenialink.enums.register import REG_ACCESS, REG_DTYPE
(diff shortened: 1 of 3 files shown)


ingeniamotion/fsoe.py
```diff
@@ -8,6 +8,7 @@
     DictionaryItem,
     DictionaryItemInputOutput,
     MasterHandler,
+    State,
```
polfeliu, line 11: Somehow ingeniamotion should also expose the module's states, right?

The fsoe_master State is not an enum. Each state is an inheritance to the State, as it has many methods for each...

If you prefer you can map it to a ingeniamotion.fsoe enum and leave fsoe_master as a completely private dependency

MartinALL91: Done

GullinBustin (approved):

polfeliu (approved):

PR #226, opened 2024-08-12
INGM-492 Ensure FSoE was configured before stop it

GullinBustin:
fsoe.stop_master get stuck if fsoe was not configured previously

Fixes # INGM-492

### Type of change

- [x] Add a flag to know if FSoE was configured

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/fsoe.py
+++ b/ingeniamotion/fsoe.py
@@ -265,6 +265,11 @@ class FSoEMasterHandler:
         """Get the FSoE master state."""
         return FSoEState(self.__master_handler.state.id)
 
+    @property
+    def running(self) -> bool:
+        """True if FSoE Master is started, else False"""
+        return self.__running
+
 
 class PDUMapper:
     """Helper class to configure the Safety PDU PDOMaps."""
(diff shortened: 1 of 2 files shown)


ingeniamotion/fsoe.py
```diff
@@ -394,6 +394,7 @@ def __init__(self, motion_controller: "MotionController") -> None:
         self.__handlers: Dict[str, FSoEMasterHandler] = {}
         self.__next_connection_id = 1
         self._error_observers: List[Callable[[FSoEError], None]] = []
+        self.__fsoe_configured = False
```
polfeliu, line 397: I think this flag should go to the respective FSoEMasterHandler.

Potentially you could do:
- create fsoe handler 1
- start fsoe handler 1
- create fsoe handler 2
- Stop all handlers. Fsoe handle 2 will be commanded to stop while it is already stopped.

Maybe with the current api it's not possible to do this but for consistency sake I would keep track of each handler

polfeliu (changes requested):

ingeniamotion/fsoe.py
GullinBustin: I keep using this property, but now I check if the master handler is running 65a7f1e

polfeliu (approved):

PR #228, opened 2024-08-21
Ingm 497 use fsoe state change callback instead of active wait

polfeliu:
[Doc]
https://novantamotion.atlassian.net/browse/INGM-497
https://docs.python.org/3/library/threading.html#event-objects

[Test]
Run fsoe example

### DIFF
--- a/ingeniamotion/fsoe.py
+++ b/ingeniamotion/fsoe.py
@@ -13,9 +13,12 @@ try:
         DictionaryItemInput,
         DictionaryItemInputOutput,
         MasterHandler,
-        State,
         StateData,
     )
+
+    if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


ingeniamotion/fsoe.py
```diff
@@ -13,6 +13,8 @@
         DictionaryItemInput,
         DictionaryItemInputOutput,
         MasterHandler,
+        State,
```
GullinBustin, line 16: State should be included inside the TYPE_CHECKING if

GullinBustin (changes requested):

MartinALL91 (approved):

GullinBustin (approved):

PR #253, opened 2024-09-19
INGM-524 Select axis 1 error

MartinALL91:
### Description

Select axis 1 error.

Fixes # INGM-524

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -83,7 +83,7 @@ class TestErrors:
         mc, alias, environment = motion_controller
         index_list = [2, 1, 3, 0]
         for i in index_list:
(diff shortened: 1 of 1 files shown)


GullinBustin (approved):

polfeliu:
Typo, It fixes INGM-524, right?

I don't understand the change though. Why should this function only work with node 1? Is there a fw bug with node 0?

MartinALL91:
> Typo, It fixes INGM-524, right?
> 
> I don't understand the change though. Why should this function only work with node 1? Is there a fw bug with node 0?
That's right, it's a FW bug with Everest. Check the linked issue for more info.

polfeliu (approved):

PR #265, opened 2024-10-23
INGM-535 Allow users to subscribe to register updates

MartinALL91:
### Description

Add methods to subscribe/unsubscribe to register value updates.

Fixes # INGM-535

### Type of change

- Add the subscribe_register_update method.
- Add the unsubscribe_register_update method.
- Append the servo alias to the information received by ingenialink.

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/communication.py
+++ b/ingeniamotion/communication.py
@@ -7,7 +7,7 @@ import zipfile
 from collections import OrderedDict
 from functools import partial
 from os import path
-from typing import TYPE_CHECKING, Any, Callable, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
 
 import ifaddr
 import ingenialogger
@@ -21,7 +21,8 @@ from ingenialink.ethercat.network import EthercatNetwork
 from ingenialink.ethernet.network import EthernetNetwork
 from ingenialink.exceptions import ILError
 from ingenialink.network import NET_DEV_EVT, SlaveInfo
-from ingenialink.servo import DictionaryFactory
+from ingenialink.register import Register
+from ingenialink.servo import DictionaryFactory, Servo
 from ingenialink.virtual.network import VirtualNetwork
 from virtual_drive.core import VirtualDrive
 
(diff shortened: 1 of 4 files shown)


ingeniamotion/communication.py
```diff
@@ -37,6 +38,47 @@
...
+
+    def __new__(
+        cls, alias: str, callback: Callable[[str, Servo, Register], None]
+    ) -> "RegisterUpdateCallbackModifier":
+        if alias in cls.instances:
+            return cls.instances[alias]
+
+        obj = super().__new__(cls)
+
+        cls.instances[alias] = obj
```
GullinBustin, line 61: With this method, you can only subscribe to one callback by alias. I'm not sure if we will need to subscribe to it more than once.

MartinALL91: I'm aware, I think I can tweak it to allow multiple callbacks per servo.

MartinALL91: Done

ingeniamotion/communication.py
```diff
@@ -37,6 +38,64 @@
...
+            "RegisterUpdateCallbackModifier",
+        ],
+    ] = {}
+
+    def __new__(
+        cls,
+        alias: str,
+        callback: Callable[[str, Servo, Register, Union[int, float, str, bytes]], None],
+    ) -> "RegisterUpdateCallbackModifier":
+        if alias in cls.instances:
```
polfeliu, line 64: Some magic going on here. 
I would explain what is going on.

Also I think that this cache mechanism is broken if you create two motioncontrollers, since the instances list is not static instead of scoped to mc

GullinBustin (approved):

ingeniamotion/communication.py
```diff
@@ -990,6 +1001,67 @@ def unsubscribe_servo_status(
...
+            if observer.im_callback == callback:
+                self.register_update_observers[drive].remove(observer)
+                break
+
+        if len(self.register_update_observers[drive]) == 0:
+            del self.register_update_observers[drive]
+            # No observers, unsubscribe from ingenialink
+            drive.register_update_unsubscribe(self._il_callback)
+
+    def _il_callback(
```
polfeliu, line 1050: I wrote this name on the mockup heh but maybe it's a bit descontextualized here

polfeliu: _il_register_subscribe_callback?

tests/test_communication.py
```diff
@@ -623,3 +639,34 @@ def test_get_available_canopen_devices(mocker):
...
+    assert register_update_callback.call_count == 2
+    assert register_update_callback.alias == alias
+    assert register_update_callback.register.identifier == user_over_voltage_uid
+    assert register_update_callback.value == new_reg_value
+
+    mc.communication.unsubscribe_register_update(
+        register_update_callback.register_update_test, servo=alias
+    )
+
+    mc.communication.set_register(user_over_voltage_uid, value=previous_reg_value, servo=alias)
```
polfeliu, line 672: Check unsubscribe works?
```
# The old value has not been propagated after unsubscribing
assert register_update_callback.value == new_reg_value
````

polfeliu (approved):

PR #273, opened 2024-11-06
INGM-537 Explicitly export ingenialink enums

MartinALL91:
### Type of change

- Explicitly export ingenialink enums.
- Add the export decorator.

### Tests
- Run mypy static type check.
- Check that no errors are found.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/enums/__init__.py
+++ b/ingeniamotion/enums/__init__.py
@@ -1,9 +1,21 @@
 from enum import Enum, IntEnum
+from typing import Type, TypeVar
 
 from ingenialink.canopen.network import CAN_BAUDRATE, CAN_DEVICE
 from ingenialink.enums.register import REG_ACCESS, REG_DTYPE
 
+T = TypeVar("T", bound=Type[Enum])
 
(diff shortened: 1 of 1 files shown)


ingeniamotion/enums/__init__.py
```diff
@@ -1,9 +1,21 @@
...
 from ingenialink.canopen.network import CAN_BAUDRATE, CAN_DEVICE
 from ingenialink.enums.register import REG_ACCESS, REG_DTYPE
 
+T = TypeVar("T", bound=Type[Enum])
 
+__all__ = ["CAN_BAUDRATE", "CAN_DEVICE", "REG_ACCESS", "REG_DTYPE"]
+
+
+def export(obj: T) -> T:
+    """Decorator use to explicitly export a class"""
```
polfeliu, line 13: If it's T is bound to enum only works for enums, not classes

MartinALL91: I know. Only enums should be defined in this module.

GullinBustin (approved):

PR #279, opened 2024-11-13
INGM-549 Add a method to set the MAC address

MartinALL91:
### Type of change

- Add methods to set/get the MAC address.
- Add a helper class to convert MAC addresses between int and string format.
- Add an optional argument to set the MAC address in the change_tcp_ip_parametes method.

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/configuration.py
+++ b/ingeniamotion/configuration.py
@@ -1,7 +1,7 @@
 import re
 from enum import IntEnum
 from os import path
-from typing import TYPE_CHECKING, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, List, Literal, Optional, Tuple, Union, overload
 
 import ingenialogger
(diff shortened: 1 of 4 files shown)


ingeniamotion/configuration.py
```diff
@@ -876,6 +929,43 @@ def restore_tcp_ip_parameters(self, servo: str = DEFAULT_SERVO) -> None:
             raise IMException("TCP IP parameters can only be restored in ethernet servos.")
         drive.restore_tcp_ip_parameters()
 
+    def get_mac_address(
+        self, servo: str = DEFAULT_SERVO, string_format: bool = False
+    ) -> Union[int, str]:
```
polfeliu, line 934: You could scope the return type with overloads: https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading

polfeliu (approved):

GullinBustin (approved):

PR #282, opened 2024-11-15
INGM-552 Fix overloads

MartinALL91:
### Type of change

- Fix overloads.

### Tests
- Run the tests contained in the Jira issue.
- Check there are no mypy errors,

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/configuration.py
+++ b/ingeniamotion/configuration.py
@@ -1,7 +1,7 @@
 import re
 from enum import IntEnum
 from os import path
-from typing import TYPE_CHECKING, List, Literal, Optional, Tuple, Union, overload
+from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 
 import ingenialogger
 from ingenialink.canopen.network import CAN_BAUDRATE, CanopenNetwork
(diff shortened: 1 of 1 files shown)


polfeliu (approved):

polfeliu (changes requested):
i have to re-review this

ingeniamotion/configuration.py
```diff
@@ -930,11 +930,15 @@ def restore_tcp_ip_parameters(self, servo: str = DEFAULT_SERVO) -> None:
...
+    def get_mac_address(self, servo: str = ..., string_format: Literal[False] = False) -> int:
         ...
 
     @overload
-    def get_mac_address(self, *, string_format: Literal[False]) -> int:
+    def get_mac_address(self, servo: str, string_format: Literal[True]) -> str:
+        ...
+
+    @overload
+    def get_mac_address(self, *, string_format: Literal[True]) -> str:
```
GullinBustin, line 941: Why servo param is different in the three cases? Servo param is always the same type

GullinBustin (changes requested):

ingeniamotion/configuration.py
polfeliu: https://mypy.readthedocs.io/en/stable/more_types.html
"The default values of a function’s arguments don’t affect its signature – only the absence or presence of a default value does. So in order to reduce redundancy, it’s possible to replace default values in overload definitions with ... as a placeholder"

I think it's not necessary to write DEFAULT_SERVO
But at least it should contain = ...
The rest of arguments should be the same signature or more concrete (literal)

I'm updating the PR so it's not stuck forever

polfeliu: Like it's now it's working with this simple typing test:
image

polfeliu: I thought this would work:
````
   @overload
    def get_mac_address(self, servo: str = ..., string_format: Literal[False] = ...) -> int:
        ...

    @overload
    def get_mac_address(self, servo: str = ..., string_format: Literal[True] = ...) -> str:
        ...
```

But: 
ingeniamotion\configuration.py:935: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]

polfeliu (approved):

PR #284, opened 2024-12-16
INGM-557 Retrieve IP-less network adapter information

MartinALL91:
### Description

Unconfigured network adapters (IP-less) were not detected by the get_interface_name_list method.

Fixes # INGM-557

### Type of change
- Get network adapter information via WMI (Windows Management Instrumentation)

### Tests
- Remove the IPv4 and IPv6 options from a network adapter options.
- Call the get_interface_name_list method
- Check that the network adapter is listed.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/communication.py
+++ b/ingeniamotion/communication.py
@@ -379,18 +379,19 @@ class Communication(metaclass=MCMetaClass):
         coco_dict = DictionaryFactory.create_dictionary(coco_dict_path, Interface.ETH)
         self.mc.servos[alias].dictionary += coco_dict
 
-    @staticmethod
-    def __get_adapter_name(index: int) -> str:
+    @classmethod
+    def __get_adapter_name(cls, index: int) -> str:
         """Returns the adapter name of an adapter based on its index.
 
         Args:
             index : position of interface selected in
                 :func:`get_interface_name_list`.
         """
-        adapter = list(ifaddr.get_adapters())[index]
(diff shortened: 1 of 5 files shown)


ingeniamotion/communication.py
```diff
@@ -453,15 +465,68 @@ def get_ifname_by_index(self, index: int) -> str:
...
+                if o.GUID is not None
+            ]:
+                if adapter not in network_adapters:
+                    network_adapters.append(adapter)
+
+            interface_name_counter = Counter(
+                [adapter.interface_name for adapter in network_adapters]
+            )
+
+            if interface_name_counter.most_common(1)[0][1] > 1:
```
polfeliu, line 518: What are those keys?

polfeliu:
How did you test it is non-admin rights tolerant?

ingeniamotion/communication.py
MartinALL91: - most_common(1): This method returns a list of the n most common elements and their counts from the most common to the least. In this case, 1 means it returns a list with the single most common element.

- [0]: This accesses the first element of the list returned by most_common(1). Since most_common(1) returns a list with one tuple, [0] gets that tuple.

- [1]: This accesses the second element of the tuple, which is the count of the most common element.

GullinBustin (approved):

polfeliu (approved):

PR #301, opened 2025-02-12
INGM-579 Replace the check_servo decorator

MartinALL91:
### Description

The check_servo decorator uses the [inspect.getfullargspec()](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec) which can be a cause of performance issues. 

The decorator is replaced with the _get_drive method.

Fixes # INGM-579

### Type of change

- Remove the check_servo decorator.
- Update modules.
- Update tests.

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/capture.py
+++ b/ingeniamotion/capture.py
@@ -152,7 +152,6 @@ class Capture:
             NotImplementedError: If an wrong monitoring version is requested.
 
         """
-        self.mc._get_drive(servo)
         version = self._check_version(servo)
         if version == MonitoringVersion.MONITORING_V3:
             return MonitoringV3(self.mc, servo)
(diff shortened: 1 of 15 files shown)


polfeliu (approved):

ingeniamotion/capture.py
```diff
@@ -151,6 +152,7 @@ def create_empty_monitoring(self, servo: str = DEFAULT_SERVO) -> Monitoring:
             NotImplementedError: If an wrong monitoring version is requested.
 
         """
+        self.mc._get_drive(servo)
```
GullinBustin, line 155: You added a `self.mc._get_drive(servo)`, but you don't need the output, which is weird. In this case, I think that `self._check_version(servo)` should crash when trying to get the drive, I don't think it's necessary to force a `self.mc._get_drive(servo)` call.

MartinALL91: Ok, I see it.

GullinBustin (approved):

PR #305, opened 2025-02-18
Ingk 1016 create empty adapters cython module and configure pipeline for compilation

julieta-prieto:
### Description

`ingenialink` has compiled files for Windows platform.
The generated wheels `ingenialink` are saved (artifacts) in the corresponding Jenkins job, and it is copied here. In that way, the Cython compilation won't run again.

⚠️ Note that jenkins job will never pass if the `ingenialink` commit in `tox.ini` does not have a jenkins job associated with it (the job **must** contain a wheel).

### Type of change

- [X] Copy `ingenialink` wheel from Jenkins build and use it for tox pytests.

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,10 +102,10 @@ pipeline {
         }
 
         stage('Get Ingenialink Build Number') {
+            when {
+                expression { env.INGENIALINK_COMMIT_HASH != null }
+            }
             steps {
-                when {
-                    expression { env.INGENIALINK_COMMIT_HASH != null }
-                }
                 script {
                     def destDir = "ingenialink_wheels/"
                     // def commitHash = '9160029940c8fa67ee6a68284bbf2e625030d114'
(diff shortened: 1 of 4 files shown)


MartinALL91 (approved):

pytest.ini
```diff
@@ -1,5 +1,6 @@
 [pytest]
 testpaths = tests
+; addopts = --import-mode=importlib
```
GullinBustin, line 3: Remove it

tests/test_communication.py
```diff
@@ -25,6 +25,17 @@
 
 from .setups.descriptors import DriveCanOpenSetup, EthernetSetup, Setup
 
+
+@pytest.fixture
```
GullinBustin, line 29: Why do you use a feature instead of add this in the test?

Jenkinsfile
```diff
@@ -13,38 +13,70 @@ DEFAULT_PYTHON_VERSION = "3.9"
...
+                    "--setup tests.setups.rack_setups.${setup_name} " +
+                    "--cov=ingeniamotion " +
+                    "--job_name=\"${env.JOB_NAME}-#${env.BUILD_NUMBER}-${setup_name}\""
+        } catch (err) {
+            unstable(message: "Tests failed")
+        } finally {
+            if (firstIteration) {
+                def coverage_stash = ".coverage_${setup_name}"
+                bat "move .coverage ${coverage_stash}"
+                junit "pytest_reports\\*.xml"
```
GullinBustin, line 72: I think this command should be executed for each python version

GullinBustin (approved):

PR #306, opened 2025-02-19
Ingm 509 - Fix wrong error received when performing motor_enable

lucia-teruel:
### Description

Fix wrong error received when performing a motor_enable.

Before: If the error took longer to be triggered than the status word to change, the retrieved error was the last one present on the buffer.

Now: Before the motor is enabled, the last error present on the buffer is saved. If then an error causes the motor_enable() to fail, the last error present on the buffer is compared with the saved one until a difference is found. This is performed with a configured timeout with its corresponding raised error.

Fixes # INGM-509

### Type of change
- [x] Modify motor_enable() to look for error change
- [x] Modify motor_enable() to include the error-raised-timeout

### Tests
- [x] Add a new test for the timeout raised error test_motor_enable_with_error_timeout_fault()
- [x] Run tests.

Tests have also been run locally with the proper setup that enables the triggering of the error.
### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/motion.py
+++ b/ingeniamotion/motion.py
@@ -11,7 +11,7 @@ from ingeniamotion.enums import GeneratorMode, OperationMode, PhasingMode, Senso
 from ingeniamotion.exceptions import IMTimeoutError
 from ingeniamotion.metaclass import DEFAULT_AXIS, DEFAULT_SERVO
 
-DEFAULT_MOTOR_ERROR_TIMEOUT = 6000
+DEFAULT_MOTOR_ERROR_TIMEOUT_S = 6
 
 
 class Motion:
@@ -130,14 +130,14 @@ class Motion:
         self,
         servo: str = DEFAULT_SERVO,
         axis: int = DEFAULT_AXIS,
-        error_timeout: int = DEFAULT_MOTOR_ERROR_TIMEOUT,
+        error_timeout: float = DEFAULT_MOTOR_ERROR_TIMEOUT_S,
     ) -> None:
         """Enable motor.
 
         Args:
             servo : servo alias to reference it. ``default`` by default.
             axis : servo axis. ``1`` by default.
-            error_timeout: Maximum wait for error update in milliseconds.
+            error_timeout: Maximum wait for error update in seconds.
 
         Raises:
             ingenialink.exceptions.ILError: If the servo cannot enable the motor.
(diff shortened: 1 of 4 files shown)


ingeniamotion/motion.py
```diff
@@ -124,25 +126,47 @@ def get_operation_mode(
...
-            )
-            error_id, _, _, error_msg = self.mc.errors.get_error_data(error_code, servo=servo)
+            timeout = float(round(error_timeout / 1000))
+            start_time = time.time()
+            error_raised = False
+            while not error_raised and (time.time() < (start_time + timeout)):
+                error_raised = (
+                    self.mc.errors.get_number_total_errors(servo=servo, axis=axis) != num_errors
+                )
+                if error_raised:
```
MartinALL91, line 159: I would move this block outside of the loop.

ingeniamotion/motion.py
```diff
@@ -3,14 +3,16 @@
...
-from ingenialink.exceptions import ILError
+from ingenialink.exceptions import ILError, ILTimeoutError
 
 if TYPE_CHECKING:
     from ingeniamotion.motion_controller import MotionController
 from ingeniamotion.enums import GeneratorMode, OperationMode, PhasingMode, SensorType
 from ingeniamotion.exceptions import IMTimeoutError
 from ingeniamotion.metaclass import DEFAULT_AXIS, DEFAULT_SERVO
 
+DEFAULT_MOTOR_ERROR_TIMEOUT = 6000
```
MartinALL91, line 14: We should specify the units.

ingeniamotion/motion.py
```diff
@@ -124,25 +126,47 @@ def get_operation_mode(
...
+        servo: str = DEFAULT_SERVO,
+        axis: int = DEFAULT_AXIS,
+        error_timeout: int = DEFAULT_MOTOR_ERROR_TIMEOUT,
+    ) -> None:
         """Enable motor.
 
         Args:
             servo : servo alias to reference it. ``default`` by default.
             axis : servo axis. ``1`` by default.
+            error_timeout: Maximum wait for error update in milliseconds.
```
MartinALL91, line 140: I think we should use seconds instead.

lucia-teruel: @MartinALL91 For me it is odd to see timeouts in seconds, when timeout values are not to be super high. But as far as I've seen in the code there are second and millisecond-timeouts, so I don't know if there is a standard...

ingeniamotion/motion.py
lucia-teruel: @MartinALL91 done!

ingeniamotion/motion.py
lucia-teruel: @MartinALL91 changed to seconds

GullinBustin (approved):

MartinALL91 (approved):

PR #313, opened 2025-03-04
Ingm 585 remove examples and tests import from ingeniamotion pytests

julieta-prieto:
### Description

`tests` and `examples` modules are referenced in the pytests, but they are not part of `ingeniamotion` package. 
This prevents from being able to use `addopts = --import-mode=importlib` in the pytests, which ensures that the package is being tested with the same wheel that will be published.

To solve this, they are dynamically imported from a local path. 
All the modules that are not part of the package should be specified in `tests/conftest.py` -> `_DYNAMIC_MODULES_IMPORT`.

Fixes # (issue)

### Type of change

- [X] Created `dynamic_loader` function to be able to import modules in pytests that are not part of `ingeniamotion` package.
- [X] Updated `README` with information on how to run the tests.

### Tests
- [ ] Run tests.

### Documentation

- [X] Use [type hints](https://docs.python.org/3/library/typing.html) for every function and argument.

### Code formatting

### Others

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -68,6 +68,7 @@ def runTestHW(markers, setup_name) {
             bat "py -${DEFAULT_PYTHON_VERSION} -m tox -e ${version} -- " +
                     "-m \"${markers}\" " +
                     "--setup tests.setups.rack_setups.${setup_name} " +
+                    "--cov " +
                     "--job_name=\"${env.JOB_NAME}-#${env.BUILD_NUMBER}-${setup_name}\""
         } catch (err) {
             unstable(message: "Tests failed")
@@ -301,6 +302,7 @@ pipeline {
                                         }
                                         bat "py -${DEFAULT_PYTHON_VERSION} -m tox -e ${RUN_PYTHON_VERSIONS} -- " +
                                                 "-m \"not ethernet and not soem and not canopen and not virtual and not soem_multislave\" "
+                                                "--cov "
                                     }
                                     post {
                                         always {
@@ -323,6 +325,7 @@ pipeline {
                                         bat "py -${DEFAULT_PYTHON_VERSION} -m tox -e ${RUN_PYTHON_VERSIONS} -- " +
                                                 "-m virtual " +
                                                 "--setup tests.setups.virtual_drive.TESTS_SETUP "
+                                                "--cov "
                                     }
                                     post {
                                         always {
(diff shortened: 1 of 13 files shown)


tests/conftest.py
```diff
@@ -11,19 +14,63 @@
 from ingeniamotion import MotionController
 from ingeniamotion.enums import SensorType
 
-from .setups.descriptors import (
+
+def dynamic_import(module_path: Union[Path, str], import_name: Optional[Union[str, list[str]]]):
+    if not isinstance(module_path, Path):
```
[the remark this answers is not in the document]
julieta-prieto: I am not sure I totally understand to what you refer when you say that it looks so complicated. I think you mean the whole logic of using the `dynamic_import` function rather than the direct import, so I'll try to explain why it is needed.

We have realized that the `tox.ini` configuration for the tests is not testing the wheel (which would be uploaded to pypi at some point), but the local cloned repo. This means that if there is a problem with the wheel generation, the tests will never pick that.

`addopts = --import-mode=importlib` allows to use the installed version of the package instead of the local one.
This means that only the modules that are contained in the package can be imported. Therefore, `tests` and `examples` are not valid modules to import from.

Since there are tests testing those examples and picking configurations from inside `tests` folders, the dynamic import (which will look for the modules that are not part of the package in the cloned repo rather than in the wheel) is needed in order to be able to activate `addopts = --import-mode=importlib`.

Maybe there's another way of doing this that I am not aware of. Please let me know if you know a better way @polfeliu

MartinALL91 (approved):

Jenkinsfile
```diff
@@ -68,7 +68,6 @@ def runTestHW(markers, setup_name) {
             bat "py -${DEFAULT_PYTHON_VERSION} -m tox -e ${version} -- " +
                     "-m \"${markers}\" " +
                     "--setup tests.setups.rack_setups.${setup_name} " +
-                    "--cov=ingeniamotion " +
```
polfeliu, line 71: This relates to https://github.com/ingeniamc/ingenialink-python/pull/550#pullrequestreview-2679740202 right?

polfeliu (approved):

Jenkinsfile
julieta-prieto: yes!

julieta-prieto: @polfeliu I updated the changes here according to the ones in INGK

MartinALL91 (approved):

PR #323, opened 2025-03-31
INGK-1076: Adapt code to accept pysoem nogil functionalities

julieta-prieto:
### Description

Expose GIL release configuration.

### Type of change

- [X] Add GIL release configuration argument where `EthercatNetwork` is created. It uses `pysoem` configuration by default (`release_gil=False`).

### Tests
- [X] Run tests.

### Documentation

- [X] Add the changes at the `[Unreleased]` section of the [CHANGELOG](CHANGELOG.md).

### Code formatting

### DIFF
--- a/ingeniamotion/communication.py
+++ b/ingeniamotion/communication.py
@@ -736,6 +736,7 @@ class Communication:
         alias: str = DEFAULT_SERVO,
         servo_status_listener: bool = False,
         net_status_listener: bool = False,
+        gil_release_config: GilReleaseConfig = GilReleaseConfig(),
     ) -> None:
         """Connect to an EtherCAT slave - CoE.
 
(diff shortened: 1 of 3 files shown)


GullinBustin (approved):

ingeniamotion/communication.py
```diff
@@ -708,7 +710,9 @@ def connect_servo_ethercat(
         if not path.isfile(dict_path):
             raise FileNotFoundError(f"Dict file {dict_path} does not exist!")
         if interface_name not in self.mc.net:
-            self.mc.net[interface_name] = EthercatNetwork(interface_name)
+            self.mc.net[interface_name] = EthercatNetwork(
+                interface_name, gil_release_config=gil_release_config
```
polfeliu, line 714: Passing a reference here to the default dataclass seems like this pitfall:
https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

On others methods it is clear that the network is created and destroyed on the same call, but not here.
Maybe you can convert the dataclass to frozen to avoid this problem altogether

PR #324, opened 2025-04-03
INGM-596 Create a SinCos encoder test treating encoder as incremental

MartinALL91:
### Description

Create a feedback test for the SinCos encoder.

Fixes # INGM-596

### Type of change

- Create the SinCos encoder feedback test.

### Tests
- [ ] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/feedbacks.py
+++ b/ingeniamotion/feedbacks.py
@@ -20,6 +20,7 @@ class Feedbacks:
         SensorType.BISSC2: SensorCategory.ABSOLUTE,
         SensorType.QEI2: SensorCategory.INCREMENTAL,
         SensorType.INTGEN: SensorCategory.ABSOLUTE,
+        SensorType.SINCOS: SensorCategory.INCREMENTAL,
     }
 
     __feedback_polarity_register_dict = {
@@ -29,6 +30,7 @@ class Feedbacks:
         SensorType.SSI2: "FBK_SSI2_POS_POLARITY",
         SensorType.BISSC2: "FBK_BISS2_POS_POLARITY",
         SensorType.QEI2: "FBK_DIGENC2_POLARITY",
+        SensorType.SINCOS: "FBK_SINCOS_POLARITY",
     }
 
     COMMUTATION_FEEDBACK_REGISTER = "COMMU_ANGLE_SENSOR"
@@ -47,6 +49,7 @@ class Feedbacks:
             SensorType.SSI2: self.get_secondary_ssi_resolution,
             SensorType.BISSC2: self.get_absolute_encoder_2_resolution,
             SensorType.QEI2: self.get_incremental_encoder_2_resolution,
+            SensorType.SINCOS: self.get_sincos_encoder_resolution,
             SensorType.INTGEN: self.__no_feedback_resolution,
         }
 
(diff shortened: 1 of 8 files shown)


ingeniamotion/wizard_tests/feedbacks_tests/sincos_encoder_test.py
```diff
@@ -0,0 +1,22 @@
...
+
+    BACKUP_REGISTERS_SINCOS: list[str] = [FEEDBACK_POLARITY_REGISTER]
+
+    SENSOR_TYPE_FEEDBACK_TEST = SensorType.SINCOS
+
+    def __init__(
+        self, mc: "MotionController", servo: str, axis: int, logger_drive_name: Optional[str] = None
+    ) -> None:
+        super().__init__(mc, servo, axis, logger_drive_name)
+        self.backup_registers_names.extend(self.BACKUP_REGISTERS_SINCOS)
```
polfeliu, line 22: https://github.com/ingeniamc/ingeniamotion/blob/0544f27ea101a6f755e530b9f3bf61469bfd80cf/ingeniamotion/wizard_tests/phase_calibration.py#L87C44-L87C60

The attribute is copying the reference to the list instead of creating a new one. So this must be extending the class list, which is very bad.
Don't know if it's an issue with other tests, but at least for here it should be fixed, creating a copy there, so that you can safely extend

tests/setups/virtual_drive.py
```diff
@@ -1,6 +1,5 @@
-from tests import dictionaries
+from ingenialink.ethernet.network import VIRTUAL_DRIVE_DICTIONARY
+
 from tests.setups.descriptors import VirtualDriveSetup
 
-TESTS_SETUP = VirtualDriveSetup(
-    ip="127.0.0.1", dictionary=dictionaries.VIRTUAL_DRIVE_XDF_PATH, port=1061
-)
+TESTS_SETUP = VirtualDriveSetup(ip="127.0.0.1", dictionary=VIRTUAL_DRIVE_DICTIONARY, port=1061)
```
polfeliu, line 5: Why this change? If we want to separate virtual drive from ingenialink shouldn't this make it harder?

polfeliu (approved):

ingeniamotion/feedbacks.py
```diff
@@ -569,6 +573,35 @@ def get_incremental_encoder_2_resolution(
...
+            "FBK_SINCOS_RESOLUTION", servo=servo, axis=axis
+        )
+        multiplier_reg_value = self.mc.communication.get_register(
+            "FBK_SINCOS_MULT_FACTOR", servo=servo, axis=axis
+        )
+        if not isinstance(resolution, int):
+            raise TypeError("Resolution value has to be an integer")
+        if not isinstance(multiplier_reg_value, int):
+            raise TypeError("Multiplier factor value has to be an integer")
+        multiplier_factor = typing.cast(int, 2**multiplier_reg_value)
```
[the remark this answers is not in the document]
MartinALL91: Without it, I get a `Returning Any from function declared to return "int" [no-any-return]`error.

GullinBustin (approved):

PR #350, opened 2025-06-10
Ingm 619 refactor pdu mapper for dynamic mapping

polfeliu:
Refactored pdu mapper for dynamic mapping. It contains basic functionality to manipulate fsoe maps, and to be able to convert maps to pdos, and pdos to maps with interleaving crcs and the particularities detailed in ETG5120 Section 5.3.3.

More broader details of the FSoE Frame can be found on ETG5100 Section 8

The API allows to fully control the layout of the maps for experts, but also covers the expected needs of a basic user, that wants to add a safety function, regardless of the mapping. It also decouples the mapping from the Handler, as it's possible to manipulate one without having a complete FSoEMasterHandler.

See more details of the needs in Safety Phase II Concept Proposal

Some part of the implementation was done in the fsoe-master repo instead of here as it was simpler and still generic

Fixes INGM-619

### Type of change

- [x] Created PDUMaps class
- [x] Methods to convert maps to pdos
- [x] Methods to convert pdos to maps
- [x] Add methods to intelligently add i/o on the map

### Tests
- [x] Added unit tests. Run with xdf v2 and xdf v3.
- [x] Verified coverage with html report. All relevant branches are covered by unit tests

### Documentation

### Code formatting

### Others

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,7 +8,7 @@ env_list = coverage, build, docs, format, type, py{39,310,311,312}, virtual
 # It should be empty in production.
 # Use full commit hash
 [develop]
-    ingenialink = {env:INGENIALINK_INSTALL_PATH:git+https://github.com/ingeniamc/ingenialink-python@f72e7feb6d73f0b232e15723c22750c4e3f00b1a}
+    ingenialink = {env:INGENIALINK_INSTALL_PATH:git+https://github.com/ingeniamc/ingenialink-python@50857d5aca876f0874ac9d5553cca5f39f812c61}
 
 
 [testenv]


ingeniamotion/fsoe.py
```diff
@@ -701,77 +740,371 @@ def running(self) -> bool:
...
+        align_to = 1
+        if element.data_type.bits > 8:
+            align_to = 8
+        if isinstance(element, FSoEDictionaryItemOutput):
+            self.outputs.insert_in_best_position(element, align_to)
+        if isinstance(element, FSoEDictionaryItemInput):
+            self.inputs.insert_in_best_position(element, align_to)
+        if isinstance(element, FSoEDictionaryItemInputOutput):
+            self.inputs.insert_in_best_position(element, align_to)
+            self.outputs.insert_in_best_position(element, align_to)
```
julieta-prieto, line 823: can we use `elif` statements and raise an error in `else` if the type is different?

ingeniamotion/fsoe.py
```diff
@@ -701,77 +740,371 @@ def running(self) -> bool:
...
+                pdo_map.data_length_bytes
+            )
+            for valid_bit in range(valid_byte * 8, (valid_byte + 1) * 8)
+        )
+        dictionary: FSoEDictionary = dictionary_map.dictionary
+        position_bits = 0
+
+        # Number of bits of variables that are in multiple data slots
+        # That will be on future virtual padding
+        pending_virtual_padding_bits = 0
```
julieta-prieto, line 909: is there any documentation regarding this?

ingeniamotion/fsoe.py
```diff
@@ -701,77 +740,371 @@ def running(self) -> bool:
...
-class PDUMapper:
+@dataclass()
+class FSoEFrameElements:
+    """FSoE Frame Elements.
+
+    Indicates uids of elements that compose the FSoE frame, excluding the safe data.
+    """
+
+    command_uid: str
+    _crcs_prefix: str
```
julieta-prieto, line 751: Why do you define crc prefix with "_"? You are still accessing the attribute from the outside right?

ingeniamotion/fsoe.py
```diff
@@ -701,77 +740,371 @@ def running(self) -> bool:
         return self.__running
 
 
-class PDUMapper:
+@dataclass()
```
julieta-prieto, line 743: could it be frozen?

ingeniamotion/fsoe.py
```diff
@@ -701,77 +740,371 @@ def running(self) -> bool:
...
+        pdo_item_type: type[PDOMapItem],
+        servo_dictionary: "Dictionary",
+    ) -> PDOMapItem:
+        try:
+            return self.__create_pdo_item(
+                servo_dictionary, frame_elements.get_crc_uid(data_slot_i), pdo_item_type
+            )
+        except KeyError as e:
+            raise NotImplementedError(
+                f"No CRC found for data slot {data_slot_i}. Probably the PDU Map is wide"
```
julieta-prieto, line 1016: I don't get this error, what do you mean with "PDU Map is wide"?

ingeniamotion/fsoe.py
```diff
@@ -701,77 +740,371 @@ def running(self) -> bool:
...
+            if item.position_bits >= slot_bit_maximum:
+                # This item must go in the next data slot
+                # Add a CRC item, and update to the next data slot
+                pdo_map.add_item(
+                    self.__get_crc_item(
+                        data_slot_i, frame_elements, pdo_item_type, servo_dictionary
+                    )
+                )
+                data_slot_i += 1
+                slot_bit_maximum += self.__SLOT_WIDHT
```
julieta-prieto, line 1055: if this does not happen:
```
if slot_bit_maximum == 8 and item.position_bits + item.bits >= slot_bit_maximum:
                # Since there's enough data to overflow the initial slot of 8 bits,
                # it will be of 16 bits instead
                slot_bit_maximum = self.__SLOT_WIDHT
```

do you still need to increase this by `self.__SLOT_WIDHT` or by 8?

tests/test_fsoe_master.py
```diff
@@ -226,3 +234,513 @@ def test_motor_enable(mc_state_data):
...
+        maps.inputs.add(fsoe_dict.name_map[STOFunction.COMMAND_UID])
+        maps.inputs.add(fsoe_dict.name_map[SS1Function.COMMAND_UID])
+        maps.inputs.add_padding(bits=6)
+        maps.inputs.add(fsoe_dict.name_map[SafeInputsFunction.SAFE_INPUTS_UID])
+        maps.inputs.add_padding(bits=7)
+
+        rpdo = RPDOMap()
+        # Registers that are present in the map,
+        # are cleared when the map is filled
+        rpdo.add_registers(safe_dict.get_register("DRV_OP_CMD"))
```
julieta-prieto, line 348: With this comment, do you mean that "DRV_OP_CMD" should not be present anymore?
If so, where are you checking that?

ingeniamotion/fsoe.py
polfeliu: by 16.
The only case when it's a multiple of 8 and not 16 is when it's 8

ingeniamotion/fsoe.py
polfeliu: No. Objects inside are meant to be mutable

ingeniamotion/fsoe.py
polfeliu: That there are too many items in the map for the support map with

ingeniamotion/fsoe.py
polfeliu: Regarding the algorithm implementation or the behaviour on the pdo map?
The algorithm I think has enough comments
The behaviour can be found on ETG5120 Section 5.3.3

ingeniamotion/fsoe.py
polfeliu: Sure

PR #361, opened 2025-06-20
Ingm 637 read pdo map from drive instead of loading phase i default

polfeliu:
### Description

When starting the pdo maps now are read from the drive, instead of loading a hard coded one from Phase I.
The mapping is deduced from the pdo and loaded into the fsoe master handler.
Since I needed to add categories and other things to be able to reconstruct the PDO correctly, I have added more info on the registers here: https://github.com/ingeniamc/ingenialink-python/pull/586
As a result there's no point in having a reader for xdfv2 and xdfv3 so I've unified them here, which makes this PR excellent in my favourite metric:
image

Fixes INGM-637

### Type of change

- [x] Read safe pdos from drive
- [x] Generalize dictionary creator for fsoe. Agnostic of xdfv2 or xdfv3

### Tests
- [x] Run tests.

### Documentation

### Code formatting

- [ ] Use the ruff package to format the code: `ruff format ingeniamotion tests`.
- [ ] Use the ruff package to lint the code: `ruff check ingeniamotion tests`.

### Others

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,7 +8,7 @@ env_list = coverage, build, docs, format, type, py{39,310,311,312}, virtual
 # It should be empty in production.
 # Use full commit hash
 [develop]
-    ingenialink = {env:INGENIALINK_INSTALL_PATH:git+https://github.com/ingeniamc/ingenialink-python@d11208ab362e547a9229a5fdebab8205f7496d11}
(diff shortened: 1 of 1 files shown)


ingeniamotion/fsoe.py
```diff
@@ -262,6 +262,14 @@ def __init__(
         if not FSOE_MASTER_INSTALLED:
             return
         self.__servo = servo
+        self.__running = False
```
polfeliu, line 265: there are many function calls on the constructor. Some expected this attribute to be present but was not ready yet. Moved all attributes here to be safer

PR #384, opened 2025-07-24
INGM-638 PDUMaps Validator

julieta-prieto:
### Description

Introduces a validator for PDU maps. The following rules should be taken into consideration when creating a ma:

* Frames may contain 1 to 8 blocks of safe data (payload), each block being word-sized (2 bytes). 
   If the frame contains only one block, it may be either 8 bits or 16 bits.
    * Rule: `FSoEFrameRules.SAFE_DATA_BLOCKS_VALID`
    * Validator: `SafeDataBlocksValidator`

* A frame can contain up to 45 objects in total. Rule to check the number items mapped and subtracting the number of fsoe frame elements used.
    * Rule: `FSoEFrameRules.OBJECTS_IN_FRAME`
    * Validator: `SafeDataBlocksValidator`

* Padding blocks may range from 1 to 16 bits.
    * Rule: `FSoEFrameRules.PADDING_BLOCKS_VALID`
    * Validator: `PaddingBlockValidator`

* 16-bit and larger objects must be word-aligned within the payload.
    * Rule: `FSoEFrameRules.OBJECTS_ALIGNED`
    * Validator: `ObjectsAlignedValidator`

* Only 32-bit objects may be split across multiple safe data blocks.
    * Rule: `FSoEFrameRules.OBJECTS_SPLIT_RESTRICTED`
    * Validator: `SafeDataBlocksValidator`

* The STO command must always be mapped to the first position in the Safe Outputs payload.
    * Rule: `FSoEFrameRules.STO_COMMAND_FIRST`
    * Validator: `STOCommandFirstValidator`

When it is possible, the exception will include the positions in bits where a rule is invalid.
The position will be `None` otherwise.

### Type of change

- [X] Created `FSoEDictionaryMapValidator` to validate PDU maps.
- [X] Updated `safety_mapping_example.py` to verify that the mapping is valid before configuring the PDOs.
- [X] Created `frame.py` and moved all methods related to the FSoE frame.

### Tests
- [X] Run tests.

### Documentation

- [X] Add the changes at the `[Unreleased]` section of the [CHANGELOG](CHANGELOG.md).

### Code formatting

### Others

- [X] Set fix version field in the Jira issue.


tests/test_fsoe_master.py
```diff
@@ -942,3 +951,301 @@ def test_insert_in_best_position(self, sample_safe_dictionary):
...
+    @pytest.mark.fsoe
+    def test_validate_safe_data_padding_blocks(self, sample_safe_dictionary):
+        """Test that PaddingBlockValidator fails when padding blocks are not 8 bits."""
+        _, fsoe_dict = sample_safe_dictionary
+        maps = PDUMaps.empty(fsoe_dict)
+
+        maps.inputs.add_padding(bits=32)
+
+        validator = FSoEDictionaryMapValidator()
+        exceptions = validator.validate_dictionary_map_fsoe_frame_rules(
```
polfeliu, line 1131: I think this content should be directly accessible through the maps object

ingeniamotion/fsoe_master/maps_validator.py
```diff
@@ -0,0 +1,515 @@
...
+            dictionary_map: The dictionary map to validate.
+            rules: Optional list of specific rules to validate. If None, all rules are validated.
+
+        Returns:
+            Dictionary of validation errors for the FSoE frame.
+        """
+        rules_to_validate = list(set(rules)) if rules is not None else list(FSoEFrameRules)
+        for rule in rules_to_validate:
+            if rule in self._validated_rules:
+                logger.warning("Validation already performed, returning cached exceptions.")
```
polfeliu, line 504: Why does the validator have a state? And a cache.
I would expect that there are methods that produce a validation output. And that this output is immutable

tests/test_fsoe_master.py
```diff
@@ -942,3 +951,301 @@ def test_insert_in_best_position(self, sample_safe_dictionary):
...
+        _, fsoe_dict = sample_safe_dictionary
+        maps = PDUMaps.empty(fsoe_dict)
+
+        # Create a map with safe data blocks that are not 16 bits
+        maps.inputs.add(fsoe_dict.name_map[self.TEST_SI_U8_UID])  # 8 bits
+
+        # Use a dummy slot width to simulate that the safe data block is wrongly sized
+        dummy_slot_width = 2
+        mocker.patch(
+            "ingeniamotion.fsoe_master.maps.PDUMaps._PDUMaps__SLOT_WIDTH", dummy_slot_width
```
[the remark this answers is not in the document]
julieta-prieto: The data slots are properly split by definition, so I don't think there is any other way of producing the failure

I modified a bit the code to be able to use the same data slot generator in `PDUMaps` and in the validator to be sure that is always right. But yes, probably it is not worth checking...

ingeniamotion/fsoe_master/maps_validator.py
julieta-prieto: So this for was actually wrong, it was from previous code and as you see it makes no sense at all, the for is doing nothing

With cache I mean previous result, we could remove `self._exceptions` attribute from all the validators and just return the exceptions

I mainly wrote them as an attribute just in case the process image wants to access them for something...

tests/test_fsoe_master.py
```diff
@@ -942,3 +951,301 @@ def test_insert_in_best_position(self, sample_safe_dictionary):
...
+        assert FSoEFrameRules.SAFE_DATA_BLOCKS_VALID in exceptions
+        exception = exceptions[FSoEFrameRules.SAFE_DATA_BLOCKS_VALID]
+        assert isinstance(exception, InvalidFSoEFrameRule)
+        assert f"Safe data block 0 must be 16 bits, found {dummy_slot_width}" in exception.exception
+        assert exception.position == [0]
+        assert validator.is_rule_valid(FSoEFrameRules.SAFE_DATA_BLOCKS_VALID) is False
+
+    @pytest.mark.fsoe
+    def test_validate_safe_data_blocks_pdu_empty(self, sample_safe_dictionary):
+        """Test that SafeDataBlocksValidator passes when no safe data blocks are present."""
```
polfeliu, line 984: If it passes shound't len(exceptions) == 0?

tests/test_fsoe_master.py
```diff
@@ -942,3 +951,301 @@ def test_insert_in_best_position(self, sample_safe_dictionary):
...
+        )
+        assert len(exceptions) == 1
+        assert FSoEFrameRules.OBJECTS_SPLIT_RESTRICTED in exceptions
+        exception = exceptions[FSoEFrameRules.OBJECTS_SPLIT_RESTRICTED]
+        assert isinstance(exception, InvalidFSoEFrameRule)
+        assert exception.exception == (
+            "Data slot 0 contains an object that is not "
+            "32 bits, it cannot be split across multiple safe data blocks"
+        )
+        assert exception.position == [0, 1, 7, 8, 9]  # All items in the first block
```
polfeliu, line 1083: Position is in bits?
Will this be practical for showing to the user of the process image editor? @lucia-teruel

tests/test_fsoe_master.py
```diff
@@ -942,3 +951,301 @@ def test_insert_in_best_position(self, sample_safe_dictionary):
...
+            validator = FSoEDictionaryMapValidator()
+            exceptions = validator.validate_dictionary_map_fsoe_frame_rules(
+                maps.inputs, rules=[FSoEFrameRules.SAFE_DATA_BLOCKS_VALID]
+            )
+            assert FSoEFrameRules.SAFE_DATA_BLOCKS_VALID not in exceptions
+            assert validator.is_rule_valid(FSoEFrameRules.SAFE_DATA_BLOCKS_VALID) is True
+
+    @pytest.mark.fsoe
+    def test_validate_safe_data_padding_blocks(self, sample_safe_dictionary):
+        """Test that PaddingBlockValidator fails when padding blocks are not 8 bits."""
```
polfeliu, line 1124: are not between 1-16 bits?

tests/test_fsoe_master.py
```diff
@@ -942,3 +951,301 @@ def test_insert_in_best_position(self, sample_safe_dictionary):
...
+        assert not validator.exceptions
+        assert FSoEFrameRules.OBJECTS_ALIGNED not in validator._validated_rules
+        validator.validate_dictionary_map_fsoe_frame_rules(
+            maps.inputs, rules=[FSoEFrameRules.OBJECTS_ALIGNED]
+        )
+        assert not validator.exceptions
+        assert validator.is_rule_valid(FSoEFrameRules.OBJECTS_ALIGNED) is True
+
+    @pytest.mark.fsoe
+    def test_validato_sto_command_first_in_outputs(self, sample_safe_dictionary):
```
polfeliu, line 1191: validator

polfeliu (changes requested):

ingeniamotion/fsoe_master/maps_validator.py
```diff
@@ -0,0 +1,510 @@
...
+class FSoEFrameRules(Enum):
+    """FSoE frame rules."""
+
+    SAFE_DATA_BLOCKS_VALID = (
+        "Frame must contain 1-8 blocks of safe data, each 16-bit (except single block may be 8-bit)"
+    )
+    OBJECTS_IN_FRAME = "A frame can contain up to 45 objects in total"
+    PADDING_BLOCKS_VALID = "Padding blocks may range from 1 to 16 bits"
+    OBJECTS_ALIGNED = "16-bit and larger objects must be word-aligned within the payload"
+    OBJECTS_SPLIT_RESTRICTED = "Only 32-bit objects may be split across multiple safe data blocks"
```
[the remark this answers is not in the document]
julieta-prieto: Do you have any better suggestions? 😅

I am supposing here that the users know that we have 16 bit words and that they should fit the data accordingly.

This message means that an object smaller than 16 bits is split across multiple blocks. I am not sure how to present this information in a clearer way. Maybe: _"Make sure that 8 bit objects belong to the same data block"_ ?

Also, I believe that it will be hard to keep track of the blocks if there are a lot of inputs/ outputs. Maybe we should think of highlighting the stuff that belongs to the same data block in a single color (background color cell).

tests/test_fsoe_master.py
julieta-prieto: docstring is wrong, it shouldn't pass

tests/test_fsoe_master.py
```diff
@@ -1063,3 +1072,298 @@ def test_insert_in_best_position(self, sample_safe_dictionary):
...
+        maps = PDUMaps.empty(fsoe_dict)
+        # Only validate the safe data blocks rule
+        validator = FSoEDictionaryMapValidator()
+        output = validator.validate_dictionary_map_fsoe_frame_rules(
+            maps.inputs, rules=[FSoEFrameRules.SAFE_DATA_BLOCKS_VALID]
+        )
+        assert len(output.exceptions) == 1
+        assert FSoEFrameRules.SAFE_DATA_BLOCKS_VALID in output.exceptions
+        exception = output.exceptions[FSoEFrameRules.SAFE_DATA_BLOCKS_VALID]
+        assert exception.exception == "No safe data blocks found in PDO map"
```
polfeliu, line 1116: STO is mandatory on the outputs, so at least you have 1 byte there for sure.

Are we sure it is not valid to have no data on the inputs?

And by no data I mean a padding of 8 bits on the pdo.
To be honest I'm not sure if __fill_pdo_map produces a 8 bit padding from a empty process image.
But maybe it should, in the same way we don't force the user to add the trailing padding to the process image, it is implicit

polfeliu:
The last part of the issue is missing but maybe it's worth to split it in another:
"Additionally we should try to make the functions like “insert_in_best_position” as user-friendly as possible, fulfilling the alignment required or any other consideration"

ingeniamotion/fsoe_master/maps_validator.py
julieta-prieto: @polfeliu I have changed this a bit, there is no more `self._exceptions` and the result can't be "cached" anymore.
`FSoEFrameRuleValidatorOutput` should contain all the information needed

tests/test_fsoe_master.py
julieta-prieto: Regarding this "Are we sure it is not valid to have no data on the inputs?", it is invalid indeed, this is what this test checks:
https://github.com/ingeniamc/ingeniamotion/blob/5f1d037f24f60617559410f79043adab4cb2e0a4/tests/test_fsoe_master.py#L1106

So the validation of all the rules at the same time will fail, but in my opinion, this individual rule shouldn't fail, as it is only mandatory for the outputs

julieta-prieto:
> The last part of the issue is missing but maybe it's worth to split it in another: "Additionally we should try to make the functions like “insert_in_best_position” as user-friendly as possible, fulfilling the alignment required or any other consideration"

created another issue @polfeliu : https://novantamotion.atlassian.net/browse/INGM-677

ingeniamotion/fsoe_master/maps_validator.py
```diff
@@ -0,0 +1,500 @@
...
+            If the safe data blocks are valid, then the number of objects in the frame
+            will always be valid as well.
+            If not, then there will be an overflow error when trying to write the PDU.
+
+        Args:
+            dictionary_map: The dictionary map to validate.
+            safe_data_blocks: The list of safe data blocks to validate.
+        """
+        n_crcs = len(safe_data_blocks)  # One CRC per safe data block
+        n_registers = len(dictionary_map._items)
```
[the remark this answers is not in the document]
polfeliu: INGK-1155

tests/test_fsoe_master.py
polfeliu: But should it really be unvalid?

julieta-prieto: I think that it should, right? 
image

polfeliu: The converter to pdo map appareantly already takes empty map into account although it did not have a test.
I would not place a restriction on the process image that is not needed. The PDO Map will be valid
https://github.com/ingeniamc/ingeniamotion/pull/392

julieta-prieto: changed

polfeliu (approved):

ingeniamotion/fsoe_master/maps_validator.py
```diff
@@ -0,0 +1,461 @@
...
+        for data_slot_i, slot_items in safe_data_blocks:
+            slot_size_bits = SafeDataBlocksValidator.__get_bits_in_data_block(slot_items)
+            # If the frame contains only one block, it may be either 8 bits or 16 bits
+            if n_safe_data_blocks == 1:
+                # It can be smaller -> it will be padded to 16 bits when creating the PDO frame
+                if slot_size_bits > 16:
+                    return {
+                        FSoEFrameRules.SAFE_DATA_BLOCKS_VALID: InvalidFSoEFrameRule(
+                            rule=FSoEFrameRules.SAFE_DATA_BLOCKS_VALID,
+                            exception="Single safe data block must be 16 bits or less "
```
lucia-teruel, line 211: What about: ?
``` Python
exception="Single safe data blocks must be 16 bits or less, with padding. Found: {slot_size_bits} bits"
```
I find the "completed with padding" a bit confusing maybe. I also like the idea of differentiating the _Found_ a bit more, plus adding the units.

If you agree this can be replicated on other msgs.

julieta-prieto: changed

PR #427, opened 2025-09-09
INGM-706 Improve subscription to PDO events

MartinALL91:
### Description

Improve the subscription to PDO events.

### Type of change

- Subscriptions to PDO events are established upon the FSoE master instantiation.

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingeniamotion/fsoe_master/handler.py
+++ b/ingeniamotion/fsoe_master/handler.py
@@ -149,6 +149,8 @@ class FSoEMasterHandler:
 
         self.__safety_master_pdu = servo.read_rpdo_map_from_slave(self.__master_map_object)
         self.__safety_slave_pdu = servo.read_tpdo_map_from_slave(self.__slave_map_object)
+        self.__safety_master_pdu.subscribe_to_process_data_event(self.get_request)
+        self.__safety_slave_pdu.subscribe_to_process_data_event(self.set_reply)
 
         map_editable = (self.__master_map_object.registers[0].access == RegAccess.RW) and (
             self.__slave_map_object.registers[0].access == RegAccess.RW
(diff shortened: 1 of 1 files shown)


ingeniamotion/fsoe_master/handler.py
```diff
@@ -358,10 +360,6 @@ def stop(self) -> None:
         self.__in_initial_reset = False
         self.__running = False
 
-        # Unsubscribe from PDO events
```
julieta-prieto, line 361: why are you removing this? 
maybe you should check this issue and the documentation it has associated https://novantamotion.atlassian.net/browse/INGM-703

we do need to unsubscribe at some point

ingeniamotion/fsoe_master/handler.py
```diff
@@ -149,6 +149,8 @@ def __init__(
 
         self.__safety_master_pdu = servo.read_rpdo_map_from_slave(self.__master_map_object)
         self.__safety_slave_pdu = servo.read_tpdo_map_from_slave(self.__slave_map_object)
+        self.__safety_master_pdu.subscribe_to_process_data_event(self.get_request)
+        self.__safety_slave_pdu.subscribe_to_process_data_event(self.set_reply)
```
julieta-prieto, line 153: is there a specific reason to move it here?

julieta-prieto (changes requested):

ingeniamotion/fsoe_master/handler.py
julieta-prieto: Have you checked the documentation I have mentioned? There is a case when there is a need to unsubscribe

julieta-prieto (approved):

PR #434, opened 2025-09-15
Ingm 703 improve fsoe master start stop

julieta-prieto:
### Description

Improve FSoE start/stop methods. 

### Type of change

- [X] Do not unassign PDO maps when stopping the master. As mentioned in https://github.com/ingeniamc/ingeniamotion/pull/430, PDO poller in safe drives needs safety PDOs to be mapped, even if the master is not active, so PDO maps cannot be removed when stopping the master.
- [X] Added `subscribe_to_process_data_events` method to `handler.py` to subscribe to safety PDO maps.
- [X] Subscribe to safety PDO maps when setting the PDO map to the slave, unsubscribe when stopping the handler.
- [X] Created method to start the master handler and remove start on first request with PDO callbacks.

### Tests
- [X] Run tests.

### Documentation

### Code formatting

### Others

- [X] Set fix version field in the Jira issue.

### DIFF
--- a/ingeniamotion/fsoe.py
+++ b/ingeniamotion/fsoe.py
@@ -102,7 +102,6 @@ class FSoEMaster:
 
         for master_handler in self._handlers.values():
             master_handler.start()
-            master_handler.subscribe_to_process_data_events()
 
         if start_pdos:
             for servo in self._handlers:
--- a/ingeniamotion/fsoe.py
+++ b/ingeniamotion/fsoe.py
@@ -101,8 +101,8 @@ class FSoEMaster:
             raise RuntimeError("FSoE master is not configured, can't start the master.")
 
         for master_handler in self._handlers.values():
-            master_handler.subscribe_to_process_data_events()
             master_handler.start()
+            master_handler.subscribe_to_process_data_events()
 
         if start_pdos:
             for servo in self._handlers:
--- a/ingeniamotion/fsoe.py
+++ b/ingeniamotion/fsoe.py
@@ -116,8 +116,6 @@ class FSoEMaster:
         if stop_pdos:
             for servo in self._handlers:
                 self.__mc.capture.pdo.stop_pdos(servo=servo)
-            if self.__fsoe_configured:
-                self._remove_pdo_maps_from_slaves()
         self.__fsoe_configured = False
 
     def sto_deactivate(self, servo: str = DEFAULT_SERVO) -> None:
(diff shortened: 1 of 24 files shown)


ingeniamotion/fsoe_master/handler.py
```diff
@@ -408,6 +429,9 @@ def set_pdo_maps_to_slave(self) -> None:
             rpdo_maps=[self.safety_master_pdu_map], tpdo_maps=[self.safety_slave_pdu_map]
         )
 
+        # Subscribe to process data events
+        self.subscribe_to_process_data_events()
```
polfeliu, line 433: this subscription lifecycle does not make sense to me.
If you start/stop/start you are not subscribed to the events.
maybe the pdo maps are already set to the slave and you are not setting them again...

julieta-prieto:
> Have you tested integration with ml?

Yes, there is a PR with the same branch name in ML3

ingeniamotion/fsoe_master/handler.py
```diff
@@ -418,9 +442,13 @@ def remove_pdo_maps_from_slave(self) -> None:
...
     def get_request(self) -> None:
-        """Set the FSoE master handler request to the Safety Master PDU PDOMap."""
+        """Set the FSoE master handler request to the Safety Master PDU PDOMap.
+
+        Raises:
+            RuntimeError: If the FSoE Master is not running.
+        """
         if not self.__running:
-            self.__start_on_first_request()
+            raise RuntimeError("FSoE Master is not running")
```
polfeliu, line 451: does the pdo thread have some kind of exception contention? This looks a bit hard.
not sure if a race condition could trigger this

polfeliu: In fact the fw does have this kind of condition. The fsoe slave starts a bit later than the ethercat pdo slave so there's a few cycles where the map is still empty

polfeliu: Just a concern, maybe we just have to handle this problem on anther layer if it ever comes.
We have talked about decoupling the fsoe to another thread, which could be a source of this. We'll just have to be extra careful

ingeniamotion/fsoe.py
```diff
@@ -314,14 +333,18 @@ def _delete_master_handler(self, servo: str = DEFAULT_SERVO) -> None:
             return
         self._handlers[servo].delete()
         del self._handlers[servo]
+        if not len(self._handlers) and self.__fsoe_configured:
+            self.__fsoe_configured = False
```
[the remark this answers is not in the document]
julieta-prieto: mm I don't know at which point it was introduced, but the idea is to let fsoe know when the PDOs have been configured.
If they are not configured, it will raise an error when starting the master.

PR #505, opened 2026-02-20
INGK-1123 Catch virtual drive monitoring V1 exception

MartinALL91:
This pull request makes a minor update to exception handling in the `ingeniamotion/capture.py` file. The change ensures that the code catches a more general `ILError` exception instead of just `ILIOError` when checking the monitoring version, and imports `ILError` accordingly. This improves robustness in error handling.

- Exception handling improvement:
  * Updated `_check_version` to catch `ILError` instead of only `ILIOError`, and imported `ILError` from `ingenialink.exceptions` to support this change. [[1]](diffhunk://#diff-fbf2a038c5d21ee547ace76454dc2fcbb5df34417c2c5b7f0ba59add232ce913L5-R5) [[2]](diffhunk://#diff-fbf2a038c5d21ee547ace76454dc2fcbb5df34417c2c5b7f0ba59add232ce913L321-R321)

### DIFF
--- a/ingeniamotion/capture.py
+++ b/ingeniamotion/capture.py
@@ -2,7 +2,7 @@ from typing import TYPE_CHECKING, ClassVar, Optional, Union
 
 import numpy as np
 from ingenialink.dictionary import SubnodeType
-from ingenialink.exceptions import ILError, ILIOError
+from ingenialink.exceptions import ILError
 from ingenialink.poller import Poller
 from numpy.typing import NDArray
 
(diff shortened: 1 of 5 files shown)


julieta-prieto (approved):

polfeliu:
no tests or update of deps? Is there already test for this?

MartinALL91:
> no tests or update of deps? Is there already test for this?

I’ll update the dependencies once they are merged into develop. The lower‑level libraries (INGK, virtual drive) already include tests, and while we could add similar tests here, I believe it would be redundant.

lucia-teruel (approved):

GullinBustin (approved):

PR #526, opened 2026-04-08
Cit 612 make multidrive setups and virtual drive also versioned

polfeliu:
Tested and unfortunately had to adapt to the api. https://github.com/ingeniamc/ingeniamotion/pull/526

### DIFF
--- a/tests/setups/rack_specifiers.py
+++ b/tests/setups/rack_specifiers.py
@@ -93,7 +93,7 @@ CAN_SETUP = SpecifierContainer({
 })
 
 
-ECAT_MULTISLAVE_SETUP = MultiRackServiceConfigSpecifier.create(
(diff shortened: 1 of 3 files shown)


tests/setups/rack_specifiers.py
```diff
@@ -93,7 +93,7 @@
 })
 
 
-ECAT_MULTISLAVE_SETUP = MultiRackServiceConfigSpecifier(
+ECAT_MULTISLAVE_SETUP = MultiRackServiceConfigSpecifier.create(
```
polfeliu, line 96: unfortunately this does break the API with the newly introduced version pattern as others. Only place where the dataclass was used directly instead of classmethod.
Another option that is worth considering on the future is to let go of dataclasses to have much finer control of the constructor.

julieta-prieto (approved):

GullinBustin (approved):

PR #541, opened 2026-04-22
MOT3-5526 Unhandled on incremental encoder 1 resolution test if resolution is set to 0

MartinALL91:
### Problem

When the feedback resolution is `0`, `DCFeedbacksResolutionTest` raises an unhandled `ZeroDivisionError` during setup because it divides by the resolution without validating it first. `DCFeedbacksPolarityTest` also lacks this validation, allowing it to operate with a meaningless resolution value.

The BLDC feedback test (`Feedbacks.feedback_setting()`) already guards against this by raising `ILConfigurationError` when resolution is 0. The DC test classes were missing the same check.

### Fix

Added a resolution-is-zero validation in `setup()` of both `DCFeedbacksResolutionTest` and `DCFeedbacksPolarityTest`, raising `ILConfigurationError` consistently with the existing BLDC guard.

Added virtual tests covering the zero-resolution case for BLDC, DC resolution, and DC polarity tests.

### DIFF
--- a/ingeniamotion/wizard_tests/feedbacks_tests/dc_feedback_polarity_test.py
+++ b/ingeniamotion/wizard_tests/feedbacks_tests/dc_feedback_polarity_test.py
@@ -2,6 +2,7 @@ from enum import IntEnum
 from typing import TYPE_CHECKING, ClassVar, Optional
 
 import ingenialogger
+from ingenialink.exceptions import ILConfigurationError
 from typing_extensions import override
 
 from ingeniamotion.enums import FeedbackPolarity, OperationMode, SensorType, SeverityLevel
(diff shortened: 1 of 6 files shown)


julieta-prieto (approved):

ingeniamotion/wizard_tests/feedbacks_tests/dc_feedback_polarity_test.py
```diff
@@ -2,6 +2,7 @@
 from typing import TYPE_CHECKING, ClassVar, Optional
 
 import ingenialogger
+from ingenialink.exceptions import ILConfigurationError
```
GullinBustin: Why do you use `ILConfigurationError`?

MartinALL91: To be consistent with the other check https://github.com/ingeniamc/ingeniamotion/blob/develop/ingeniamotion/wizard_tests/feedbacks_tests/feedback_test.py#L213C19-L213C23

polfeliu (approved):

GullinBustin (approved):

PR #577, opened 2026-06-03
INGK-1271 try to make reg.identifier mandatory

polfeliu:
Update to ingenialink that has reg.identifier mandatory and remove some is None checks that are unnecessary now


pyproject.toml
```diff
@@ -20,7 +20,7 @@ classifiers = [
 requires-python = ">=3.9"
 
 dependencies = [
-    "ingenialink>=7.6.0, < 8.0.0",
+    "ingenialink==7.6.1.post52+ingk.1271.try.to.774f0ba",
```
MartinALL91: Is this really necessary?

julieta-prieto (approved):

MartinALL91 (approved):

GullinBustin (approved):

pyproject.toml
polfeliu: Addressed in commit e01aaa74 ("rollback dep requirement change"): the published `dependencies` constraint is back to `ingenialink>=7.6.0, < 8.0.0`. Only the tests group still pins the branch build, which follows the existing repo pattern for unreleased ingenialink changes.

Heads-up on the failing Jenkins build ("This commit cannot be built"): `poetry check` fails with "pyproject.toml changed significantly since poetry.lock was last generated" — the rollback commit edited the main dependency constraint but poetry.lock's content-hash was not regenerated. It needs a `poetry lock` (resolved versions are unchanged since the tests-group pin still satisfies both constraints, so only the hash updates). I could not run it from my environment because the novanta package index is not reachable here.

lucia-teruel (approved):

PR #583, opened 2026-06-10
INGK-1259 Make drive context manager chainable or inheritable from a drive state

polfeliu:
Tests and that refactor of summit testing framework doesn't break ingeniamotion


julieta-prieto (approved):

GullinBustin (approved):

polfeliu:
was a check, technically not necessary

PR #585, opened 2026-06-11
Mot3 5528 review stopable test pattern

polfeliu:
This issue was for me to learn and review how the stoppable pattern works and review that it doesnt have any flaw, as we've had reports of wizard tests not being stopped or taking too much.

The pattern is mostly good but there are times when there's big gaps between opportunities to stop the test. This pull request introduces logic on the stoppable module so this can be inspected in terms of timing, and introduces a new stoppable_trace_recorder fixtures on tests that run any basetest and collect the timing information and callsites.

The fixture records and analyzes the gaps between oportunities to stop, and has a high threshold for pass (>5 seconds, STOPPABLE_GAP_THRESHOLD_SECONDS). If we modify the code in any way that introduces a big gap this will notice and trigger a warning. Additionally it also creates a markdown report with the information it has analyzed so that we can see if we are close to the threshold. 

image
image

Jenkins displays raw markdown, if you are interested in formatting there's this [plugin](https://chromewebstore.google.com/detail/markdown-viewer/ckkdlimhmcjmikdlpkmbgfkaikojcbjk)
that can be configured with this regex expression for jenkins site:
```
\.(?:markdown|mdown|mkdn|md|mkd|mdwn|mdtxt|mdtext)(?:/.*|\?.*|#.*)?$
```
image

I have created issues to:
- MOT3-5606: Check if stoppable can avoid being a singleton (minor code sanity, not shure how doable this is)
- Lower the threshold and adapt motion tests so there are opportunities more often (INGM-768)
- Propagate this pattern on INGW, that also uses basetest for all tests and also should have checking of the stop opportunities


tests/stoppable_opportunities_timing.py
julieta-prieto, line 1: Aren't you going to move this file to the framework?
I thought we agreed on moving it to avoid maintaining code on different repos

julieta-prieto (approved):

lucia-teruel (approved):

PR #586, opened 2026-06-11
Fix power cycle in tests

julieta-prieto:
Set wait for drives to false to fix wrong reconnection in power cycle

### DIFF
--- a/tests/fsoe/test_safety_errors.py
+++ b/tests/fsoe/test_safety_errors.py
@@ -65,7 +65,6 @@ def test_get_error_with_id_not_in_dict() -> None:
     assert error.error_description == "Unknown error 4660 / 0x1234"
 
 
-@pytest.mark.skip(reason="Flaky test due to the power cycle")
 @pytest.mark.fsoe_phase2
 def test_no_errors(
     mcu_error_queue_a: "ServoErrorQueue",
@@ -73,7 +72,7 @@ def test_no_errors(
 ) -> None:
     """Test methods when there are no errors"""
     # Clear any existing errors by power cycling
-    environment.power_cycle(wait_for_drives=True, reconnect_drives=True)
+    environment.power_cycle(wait_for_drives=False, reconnect_drives=True, reconnect_timeout=30)
 
     assert mcu_error_queue_a.get_number_total_errors() == 0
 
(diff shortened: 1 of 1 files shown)


MartinALL91 (approved):

tests/fsoe/test_safety_errors.py
```diff
@@ -65,15 +65,14 @@ def test_get_error_with_id_not_in_dict() -> None:
...
-@pytest.mark.skip(reason="Flaky test due to the power cycle")
 @pytest.mark.fsoe_phase2
 def test_no_errors(
     mcu_error_queue_a: "ServoErrorQueue",
     environment: "DriveEnvironmentController",
 ) -> None:
     """Test methods when there are no errors"""
     # Clear any existing errors by power cycling
-    environment.power_cycle(wait_for_drives=True, reconnect_drives=True)
+    environment.power_cycle(wait_for_drives=False, reconnect_drives=True, reconnect_timeout=30)
```
polfeliu, line 75: if you dont wait for drives how do you guarantee that you will be able to read the total number of errors?

GullinBustin: I think this is a naming issue; it actually waits for drives. We should fix the environment.power_cycle function because wait_for_drives=True is buggy.

julieta-prieto: @polfeliu  It is waiting for the drives. As you can see it has `reconnect_drives` to `True` (and a realistic timeout, default one is 0.5s).  
`wait_for_drives` is the rack service method we used a while ago to wait for the drives to be available again.

With the introduction of `reconnect_drives` method, this one is not really needed anymore.
What we do need is the reconnect method, since it is the one that rebuilds the aliases and that stuff. 

For example, in this ingenialink commit https://github.com/ingeniamc/ingenialink-python/commit/97b94a2251d885464164b10679529e1c722bf331 you can see how it was used before the update.
We used `wait_for_drives=False` with `wait_until_alive` method. In this ingeniamotion commit https://github.com/ingeniamc/ingeniamotion/commit/533351ad60d5bbb8679e910dafa407cfd42868cf I removed the recover method, but I should have set `wait_for_drives=False`.

As @GullinBustin mentions, I think that the safest solution would be to completely remove the `wait_for_drives` method from the framework. Even if it is an option that the rack provides, it no longer makes sense with the recover method that we already embedded in the framework's power cycle method.
If we want to keep it for compatibility, I would raise an error if it is `True` and `reconnect_drives` is `False`.

polfeliu: 100% agree, naming/api problem. Create issue to fix it please, i would try to mantain retrocompatibility

polfeliu (approved):

lucia-teruel (approved):

GullinBustin (approved):

PR #602, opened 2026-06-23
INGW-701 Use drive context manager for wizard tests

polfeliu:
Refactors the register backup and restore mechanism in the test framework to use a context manager for managing drive state during tests. It removes manual backup/restore logic and updates all relevant test classes to support this new approach, instead of listing registers that must be backed up and then restored, tests now list which registers are accepted to be changed after the test (blacklist -> whitelist)

**register state management**
Replaces manual register backup/restore in `BaseTest` and all derived test classes with `DriveContextManager`, simplifying code and ensuring consistent drive state handling. The `run()` method in `BaseTest` now uses this context manager to automatically restore drive registers after test execution, except for those explicitly allowed to remain changed. (`ingeniamotion/wizard_tests/base_test.py`. All registers are read at the beginning of the test and the changed ones are reverted after (the initial read will be removed when motionlab passes it directly on the future)

**Exclusions**
I've found out that some test do actually leave registers modified, this is to be discussed, and I would like it to be more controlled in some API, for now this is done with the ACCEPTED_CHANGED_REGISTERS list (a whitelist)

**Brake test enhancements**
Adapted brake in a rought way to not break it, it is quite special since its splits the workflow in two steps, and it kind of evades the control of the drive context manager. To be refactored on another issue (requires ML3 changes) MOT3-5631

**Testing**
To verify all registers are rolled back correctly, registers are read back at the end of the unit tests, and only accepted ones can change


ingeniamotion/wizard_tests/phase_calibration.py
```diff
@@ -38,26 +38,16 @@ class Phasing(BaseTest[LegacyDictReportType]):
...
-        "COMMU_PHASING_TIMEOUT",
-        "COMMU_PHASING_ACCURACY",
-        "COMMU_PHASING_MODE",
-        "MOT_COMMU_MOD",
-        "COMMU_ANGLE_INTEGRITY1_OPTION",
-        "COMMU_ANGLE_INTEGRITY2_OPTION",
-    ]
+    # Left changed on purpose: drive-computed calibration results (angle offsets) and the
+    # phasing config the test applies. They are not rolled back, so restore checks accept them.
+    ACCEPTED_CHANGED_REGISTERS: ClassVar[tuple[str, ...]] = (
```
GullinBustin, line 45: I think these registers should be always restored as well. I think test must always restore the registers, after that the user can apply the suggested register if necessary. In this test the suggested registers are definer here: https://github.com/ingeniamc/ingeniamotion/pull/602/changes#diff-918f15f7e5e09c2a4e32ae18f93959d32949f59482231fd05c9668b8b673b983R331

ingeniamotion/wizard_tests/base_test.py
```diff
@@ -42,48 +45,26 @@ class ReportBase:
 class BaseTest(ABC, Stoppable, Generic[T]):
     """Abstract base Test class."""
 
-    WARNING_BIT_MASK = 0x0FFFFFFF
+    ACCEPTED_CHANGED_REGISTERS: ClassVar[tuple[str, ...]] = ()
```
GullinBustin, line 48: It’s a good approach, but I think it should never be used. I think the test should never make changes, only propose them. What do you think?

GullinBustin:
I think you left a test accidentally, this is a new one: https://github.com/ingeniamc/ingeniamotion/blob/develop/ingeniamotion/wizard_tests/dynamic_forced_phasing.py

ingeniamotion/wizard_tests/base_test.py
julieta-prieto: what do you mean by propose but not make changes @GullinBustin ?

tests/test_all_drive_tests.py
```diff
@@ -58,13 +65,48 @@ def feedback_test_setup(
...
 @pytest.mark.ethernet
 @pytest.mark.soem
 @pytest.mark.canopen
 @pytest.mark.usefixtures("feedback_test_setup")
 # https://novantamotion.atlassian.net/browse/INGM-782
 @pytest.mark.not_valid_for_product(part_number="CAP-XCR-E")
-def test_digital_halls_test(mc, alias, feedback_list):
+def test_digital_halls_test(
+    mc, alias, feedback_list, servo: Servo, registers_baseline: DriveRegistersValue
+):
```
julieta-prieto: have you tested if this works now? it has a `not_valid_for_product` marker

tests/test_all_drive_tests.py
julieta-prieto, line 1: You have made a lot of changes in tests that are not running right now.
Have you tried them?

There could be some logic error related to the context manager or something that we can't see because they never run right?
I think I would make a quick commit commenting those markers just to see if the test pass (or fail because flaky), but there is no other possible issue that we may be introducing with these changes.

.gitignore
```diff
@@ -81,4 +81,7 @@ resources/
 ingeniamotion/_version.py
 
 # Test outputs 
-tests/outputs/
\ No newline at end of file
+tests/outputs/
+
+# Pipenv and venv
+/.venv*
```
julieta-prieto, line 87: I have already seen this in other PRs that you have made. 
Why do you need to add this? Pipenv? 😅 
It works for me without this and it has always worked. I wonder why this is necessary now.

ingeniamotion/wizard_tests/base_test.py
GullinBustin: Nowadays, the output of these tests is a dictionary with the following format:

```
{
    # (int) Result code
    "result_severity": 0,
    # (dict) Suggested register values
    "suggested_registers":
        {"COMMU_ANGLE_OFFSET": 0.12},
    # (str) Human readable result message
    "result_message": "Phasing process finished successfully"
}
```

In "suggested_registers", the test sends a list of registers with suggested values. When the test finishes, the drive returns to its initial state. Then, the user (or ML3) can choose whether to apply these suggested changes. With this approach, the user knows what changes would be made to the drive.

polfeliu: i agree, in the future i think we should change it to a proposal system that is visible to the user. but we need to analyze it with test by test basis

polfeliu: there are some cases where the change is not actively done by the test, the drive just sets the commutation offset after the phasing... So it mean rolling it back and proposing the value to the user, this deserves a good design

ingeniamotion/wizard_tests/phase_calibration.py
GullinBustin: Wrong link. The good one -> https://github.com/ingeniamc/ingeniamotion/blob/c91d0fb9ef2bc3785cace62db21a0aeb52d2107e/ingeniamotion/wizard_tests/phase_calibration.py#L331

tests/test_all_drive_tests.py
polfeliu: they are not running on the cap setup but they are running on others right?

image

polfeliu: i have only done changes with regard to the backup of the registers, i dont think there are differences between setups in terms of that

.gitignore
polfeliu: We have some projects that have the venv in project option set, others dont, i have decided it to set it globally for my poetry too, because i have started to use git worktrees and its much more easy to destroy the venv when you are done

ingeniamotion/wizard_tests/phase_calibration.py
polfeliu: they were not on the backup registers list, I'm not changing the behaviour for now. the point of this refactor is too highlight them

ingeniamotion/wizard_tests/base_test.py
polfeliu: https://novantamotion.atlassian.net/browse/MOT3-5632

polfeliu:
> I think you left a test accidentally, this is a new one: https://github.com/ingeniamc/ingeniamotion/blob/develop/ingeniamotion/wizard_tests/dynamic_forced_phasing.py

Its right here: https://github.com/ingeniamc/ingeniamotion/pull/602/changes#diff-ba1ca8e7fb25789509421b5086c5aed41ed209c40bd298ff2695f72ad6da1bc8L108

No accepted registers for that case :)

GullinBustin (approved):

julieta-prieto (approved):

MartinALL91 (approved):

PR #607, opened 2026-07-01
INGM-792 ILRegisterAccessError bad constructor usage undetected

polfeliu:
https://github.com/ingeniamc/ingeniamotion/pull/607/changes#diff-fcbde3c3d82bd1dcdfb33993bada0c78eb690048040c22b7ecf9adc1473bce29L166

motor enable was re-creating exceptions to add information about the last error of the error queue. It was extracting the type and calling the constructor, assuming it didn't have arguments. This is wrong and I don't even know how to detect it with mypy, but we should not do this. I'm not sure how to modify the exception in py 3.9 so it includes more information, but this is a perfect fit for Exception notes, which were added in 3.11, and allow to add a bit of context in each unrolling layer of the exception stack.

Since python 3.9 is already deprecated and 3.10 will be deprecated on october, i have taken the simple route of making this nicety of adding information of the last error on the queue exclusive to >= py 3.11. The guards will be removable once we drop support for those pythons.

### DIFF
--- a/ingeniamotion/motion.py
+++ b/ingeniamotion/motion.py
@@ -1,6 +1,7 @@
 import time
 from collections.abc import Generator
 from typing import TYPE_CHECKING, Optional, Union
+import sys
 
 import ingenialogger
 from ingenialink.exceptions import ILError, ILTimeoutError
@@ -10,6 +11,7 @@ if TYPE_CHECKING:
 from ingeniamotion.enums import GeneratorMode, OperationMode, PhasingMode, SensorType
 from ingeniamotion.exceptions import IMTimeoutError
 from ingeniamotion.metaclass import DEFAULT_AXIS, DEFAULT_SERVO
+
 
 DEFAULT_MOTOR_ERROR_TIMEOUT_S = 6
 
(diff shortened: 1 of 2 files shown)


ingeniamotion/motion.py
```diff
@@ -161,10 +162,14 @@ def motor_enable(
                     servo=servo, axis=axis
                 )
                 _error_id, _, _, error_msg = self.mc.errors.get_error_data(error_code, servo=servo)
+                if sys.version_info >= (3, 11):
```
GullinBustin, line 165: I think this changes the current behavior for python version < 3.11. Maybe we should leave the old implementation for these cases.

tests/test_motion.py
```diff
@@ -128,25 +139,42 @@ def test_motor_enable_with_fault(
...
         with pytest.raises(exception_type) as excinfo:
             mc.motion.motor_enable(servo=alias)
-    assert str(excinfo.value) == message
+    assert str(excinfo.value) == (
+        "The subnode 1 could not be enabled within 1000 ms. "
+        "The current subnode state is ServoState.FAULT"
+    )
+
+    if sys.version_info >= (3, 11):
+        assert excinfo.value.__notes__[0] == note
```
julieta-prieto, line 148: Is this really needed? 
For example, we discard all renovate prs for which we need to make small changes for a certain python version.

Is this an improvement, or this won't work at all for python>=3.11?

ingeniamotion/motion.py
polfeliu: i think type(e).__init__() is evil. its untyped, and very hard to catch, whcih is the reason this error existed.

I have searched ways to prohibit it with static analysis, but its hard.
I can think of ways to partially prohibit it with ast-grep, which is a system that we should adopt, it's very useful for creating specific antipattern rules. You can prohibit:

type(*).(*)

but not
cls = type(*)
cls(*)

polfeliu: i don't think this feature is that important, it's a nicety.
But most importantly leaving the older implementation also means leaving the bug

tests/test_motion.py
polfeliu: the test runs with several python versions, so this assert must only be done with python>=3.11, older pythons dont have the note. I have run the jenkins pipeline with all python versions BTW

julieta-prieto (approved):

GullinBustin (approved):

PR #613, opened 2026-07-08
Ingm 778 fix test move position

julieta-prieto:
After a full analysis, the cause of `test_move_position` failure was an invalid CRC message from the absolute encoder with BiSS-C protocol. This is a known issue that is supposed to be fixed in the upcoming 2.11.0 release.

A new test file was developed to verify this issue (see commit https://github.com/ingeniamc/ingeniamotion/pull/613/changes/4d5d5abaf1da2b766ed8fd4949a5dbbad791e199), but it has been removed since it is a known issue and we shouldn't need it.

Fixes:
* Added a new `biss_c_flaky` to indicate the tests that may fail due to this cause.
* `conftest.py` modifies the pytest collection by deselecting the tests marked as `biss_c_flaky` if a configuration file is provided and some of the feedbacks are using absolute encoder 1 or 2.
* Added a new specifier to run on SIRIUS setup, configuring the encoder with SSI protocol to make sure that the test passes even if it is not tested with BiSS-C protocol right now.
* Updated `summit_drives_ci_config` reference to include "No phasing" fix


tests/conftest.py
```diff
@@ -215,3 +312,75 @@ def refresh_registers_for_test_rollback(servo: Servo, register_uids: list[str]):
...
+
+        Raises:
+            ValueError: If the dictionary does not contain the required keys.
+        """
+        if "protocol" not in data or "resolution" not in data:
+            raise ValueError("Dictionary must contain 'protocol' and 'resolution' keys.")
+        return cls(protocol=data.get("protocol"), resolution=data.get("resolution"))
+
+
+@pytest.fixture(autouse=True, scope="session")
```
julieta-prieto, line 377: autouse will be removed once it is part of the environment: https://novantamotion.atlassian.net/browse/CIT-761

tests/setups/rack_specifiers.py
```diff
@@ -301,3 +323,48 @@
...
+                        stage_name="SIRIUS EVS-NET-E Tests - FW. 2.10.0",
+                    )
+                },
+                "configure_encoder_protocol": {"protocol": "SSI1", "resolution": 17},
+            },
+        ),
+        # BiSS-C tests are flaky on 2.10.0 should pass on 2.11.0
+        "2.11.0.005": VersionConfig.from_files(
+            version="2.11.0.005",
+            firmware=dist_path(
```
MartinALL91, line 350: Since this configuration is only used on Linux, could we use the correct Linux paths directly and remove the dist_path() helper?

MartinALL91 (approved):

GullinBustin (approved):

tests/setups/rack_specifiers.py
```diff
@@ -11,6 +14,25 @@
 
 import summit_drives_ci_configs.config_files as config_files
 
+
+def dist_path(p: str) -> Path:
```
polfeliu, line 18: im not sure if @gonimart ever create the issue to migrate this

tests/conftest.py
```diff
@@ -66,6 +79,84 @@ def pytest_configure(config):  # noqa: ARG001
...
+    for item in items:
+        if not item.get_closest_marker(__BISS_C_CONFIG_MARKER):
+            continue
+
+        for skip_product in ["CAP-*", "EVE-*", "EVS-*"]:
+            item.add_marker(
+                pytest.mark.valid_versions_for_product(part_number=skip_product, max="2.6.0")
+            )
+            item.add_marker(
+                pytest.mark.valid_versions_for_product(part_number=skip_product, min="2.11.0")
```
polfeliu, line 164: this is related to this right? https://github.com/CeleraMotion/summit-testing-framework/pull/78/changes

I didn't understand it properly I think, but it reads weird here.
If valid versions are max = "2.6.0" and min ="2.11.0" i would expect no version is valid.

I think it would be clearer to extend to have a marker of not_valid_version_for_product(EVE, min="2.6.0", max="2.11.0"),

you can even have several markers on the same test, in case a bug is fixed, reintroduced, fixed, reintroduced (unideal xd)

tests/setups/rack_specifiers.py
julieta-prieto: this particular configuration is just used in Linux, but we have used safety release candidates before, so I guess it is not impossible to use dist path in Windows in the future

tests/conftest.py
julieta-prieto: @polfeliu I created the [issue](https://novantamotion.atlassian.net/browse/CIT-763) and added it as a comment in the code

polfeliu (approved):

lucia-teruel (approved):

PR #615, opened 2026-07-09
Ingm 795 fix test motor enable with fault

julieta-prieto:
# Description

Fix motor_enable fault detection when the error counter does not advance.

This change fixes a bug in `motion.py` where `motor_enable` was treating a real drive fault as a timeout whenever it waited only on `DRV_DIAG_ERROR_TOTAL` to change. On the affected hardware, the over-temperature fault `0x00004303` was already present in the error buffer, but the total error count sometimes didn't increase. In that case, the old logic raised `ILTimeoutError` even though the real fault was already available.

The fix makes `motor_enable` check the error buffer directly while it waits, so it can report the actual fault message instead of misreporting a timeout. True timeouts still raise `ILTimeoutError` when no error code appears within the deadline.

# Tests
Validated on hardware with repeated runs of `test_motor_enable_with_fault`. The over-temperature case now resolves as the actual fault when the error is already visible, instead of being reported as a timeout.

Check [pipeline](http://jenkins.ingenia/job/Novanta%20Motion%20-%20Ingenia%20-%20Git/job/ingeniamotion/view/change-requests/job/PR-615/6/stages/?selected-node=577) with 30 runs for `test_motion.py` module.


tests/test_motion.py
```diff
@@ -167,6 +173,13 @@ def test_motor_enable_with_delayed_fault(
     patch_get_number_total_errors.side_effect = delayed_function_return(
         4, num_errors_before_test, num_errors_before_test + 1
     )
+    patch_get_last_buffer_error = mocker.patch("ingeniamotion.errors.Errors.get_last_buffer_error")
```
julieta-prieto, line 200: Before we had enough mocking total number of errors method. 
Now, `motor_enable` has been improved to also check last buffer error, so we need to mock that method as well.

ingeniamotion/motion.py
```diff
@@ -150,25 +150,31 @@ def motor_enable(
...
+                    )
+            if error_code == 0:
                 raise ILTimeoutError("Error trigger timeout exceeded.")
-
+            _error_id, _, _, error_msg = self.mc.errors.get_error_data(error_code, servo=servo)
+            if sys.version_info >= (3, 11):
+                # Adds a note to the exception with the error last error message from the queues
+                # Only available in Python 3.11+ (add_note method)
+                e.add_note(f"Error message: {error_msg}")
+            if type(e) is ILTimeoutError:
```
polfeliu, line 191: what is the purpose of this?

julieta-prieto: If `drive.enable()` raised a timeout exception, `e` is `ILTimeoutError`.
This timeout error in `enable()` method is triggered because one of the state transitions did not happen in time due to some possible reason. This reason could be a fault.

Because of this, in `motor_enable()` we check for possible errors, and this is how we find the original cause of `ILTimeoutError`. If there has been a real fault, we want to report the real fault, so we convert that specific case into `ILError(error_msg)`.

GullinBustin: Why do you use `type(e) is ILTimeoutError` instead of `isinstance(e, ILTimeoutError)` ?

julieta-prieto: That was a copilot suggestion in case errors change. I had your suggestion before, but it was changed by copilot because it could bee "too broad", `isinstance(e, ILTimeoutError)` could also contain subclasses that we may not want to catch in this specific point.

I think we don't have any right now, but it seemed correct to keep it as a defensive measure

polfeliu: I understand the implementation but not the reason. If there has been a timeout on a transition of the motor enable we want to report it. The add_note already adds the context of why the transition failed in the first place

ingeniamotion/motion.py
```diff
@@ -150,25 +150,31 @@ def motor_enable(
...
-                    self.mc.errors.get_number_total_errors(servo=servo, axis=axis) != num_errors
+            deadline = time.time() + error_timeout
+            error_code = 0
+            while time.time() < deadline and error_code == 0:
+                current_error_code, _subnode, _warning = self.mc.errors.get_last_buffer_error(
+                    servo=servo, axis=axis
                 )
-            if error_raised:
-                error_code, _subnode, _warning = self.mc.errors.get_last_buffer_error(
+                if current_error_code != 0:
```
GullinBustin: current_error_code could be different from 0 if an error exists before the motor is enabled, but this error won’t be the correct one. I think that current_error_code should be compared with the get_last_buffer_error registered before attempting to enable the motor.

ingeniamotion/motion.py
julieta-prieto: @polfeliu what I mean is that the timeout happened because of another error.
Do you mean that we want to report a timeout instead of the real error source? If this is the case, I am not completely sure our tests match this behavior.

ingeniamotion/motion.py
julieta-prieto: @GullinBustin as discussed, I implemented the change here: https://github.com/ingeniamc/ingeniamotion/pull/615/commits/830093297315bd43f1c8cf64cb7cb5a21449b07a.

We read the last error before `enable()` and we only trigger it if it is different from the previous error. This will only happen if the error counter is not updated (which happens very few times), so we are more or less in the same situation that we were in develop.

GullinBustin (approved):

ingeniamotion/motion.py
polfeliu: yes. I a ILTimeoutError is much more concrete that ILError. I dont see a reason to redefine exceptions that were already thrown, specially when we can use the add_note.

Feel free to change the test  if it doesn't match the behaviour.

tests/test_motion.py
polfeliu: BTW, if we are mocking all error functions.. does it make any sense to run this test on ethercat soem and canopen? A virtual drive test would make a lot more sense, specially if we implemented actual error queue simulation (there's a long standing issue)

ingeniamotion/motion.py
lucia-teruel: Is this case being checked on a test?

lucia-teruel: > Is this case being checked on a test?

NVM of course it is! PR is meant to fix the test :)

lucia-teruel (approved):

PR #618, opened 2026-07-10
Ingm 788 fix test start stop master

julieta-prieto:
# Description
This PR fixes the restart in `test_start_stop_master` when the FSoE master is stopped with PDOs still running and then started again. 

The failure appeared on restart, where the PDO thread could briefly feed the master stale startup bytes from the previous session and the handshake would get stuck or reset.

Debugging evidence:
```
REPLY_TRACE index=1 state=None reply=2a0000c42d0000
...
REPLY_TRACE index=50 state=<class 'fsoe_master.fsoe_master.StateReset'> reply=528800bca30100
REPLY_TRACE index=1 state=None reply=2a050025ae0000
REPLY_TRACE index=2 state=<class 'fsoe_master.fsoe_master.StateSession'> reply=2a050025ae0000
REPLY_TRACE index=3 state=<class 'fsoe_master.fsoe_master.StateSession'> reply=2a0000c42d0000
ERROR_TRACE state=<class 'fsoe_master.fsoe_master.StateSession'> reply_index=3 ... transition=SESSION_RESET1 ...
ERROR_TRACE state=<class 'fsoe_master.fsoe_master.StateSession'> reply_index=4 ... transition=SESSION_STAY2 error=Invalid slave CRC
```

There is a repeated pattern for the entire slave reply PDU for the FSoE session startup. In the failing run we saw the first post-restart replies repeat bytes from the previous session instead of advancing cleanly:
* first restart reply: `2a050025ae0000`
* then again: `2a050025ae0000`
* then the older-looking `2a0000c42d0000`

The PDO receive path is delivering startup-sized FSoE frames whose content matches earlier handshake frames, including frames that belong to the previous session.

# Fix
Mark the restart startup phase explicitly and only treats that phase as special after a previous successful `DATA` state.
During that restart startup window, it ignores only replayed startup replies that were already seen in a prior successful startup, so stale PDO data cannot restart the handshake with a bad frame.

 That keeps the normal startup path untouched, avoids hiding real errors later in `DATA`, and fixes the actual stale-frame restart issue instead of adding more delay or suppressing all handshake errors.

# Tests
[...]

PR #628, opened 2026-07-22
Ingm 788 fix startup with stale commands

polfeliu:
When an FSoE master (re)connects to a novanta fsoe slave it will respond with a reply from a previous session — e.g. a frozen safety PDU after a master replacement or a `stop()`/`start()` restart — the master reacts with transient, expected errors (`SESSION_RESET1`, `Invalid slave CRC`) that resolve on their own once the slave catches up. These were forwarded to the user's error callback as if they were real failures.

`FSoEMasterHandler` now keeps the first reply seen since `start()` as a baseline: errors triggered while the slave is repeating that baseline are logged at debug level instead of being reported. Errors raised while the handler is intentionally stopping (teardown noise such as `DATA_WD`) are suppressed the same way.

This was partially solved earlier by withholding replies whose first byte is 0 (not a valid FSoE command, i.e. the slave not ready yet), and later refactored in [#617](https://github.com/ingeniamc/ingeniamotion/pull/617) to suppress errors until the master exits the Reset state. That window is not enough: when the slave is dead or echoing stale frames, the master cycles from Reset to Session continuously, so it repeatedly leaves Reset and the spurious errors are reported again. This PR replaces the state-based window with a baseline-reply comparison that holds for as long as the slave keeps repeating the same stale answer.

Supersedes [#618](https://github.com/ingeniamc/ingeniamotion/pull/618), which also fixed this problem and a potential other one. But the implementation didn't work for the cross-master scenario, and now we cannot reproduce the scenario with real hardware 

### Type of change

- [x] Bug fix: stale-reply and teardown errors are no longer reported to the user callback.
- [x] Test infrastructure: `tests/fsoe/utils/fsoe_emulation.py` adds a software FSoE slave (`FSoESlave`) and medium (`FSoENetwork`) that drive the real master handshake to Data without hardware, including re-delivery of stale replies across a master replacement. Exchange traces are written to the summit-testing-framework output directory.
- [x] Re-enabled `test_start_stop_master` for DEN-S-NET-E (removed the `not_valid_for_product` marker).

### Tests

New hardware-less tests in `tests/fsoe/test_handler.py`:

- `test_mock_slave_drives_real_handshake_to_data_state`: the emulated slave brings the real master from Reset to Data with no errors.
- `test_master_ignores_stale_reply_after_master_replacement`: a second master taking over the same slave receives the previous session's DATA reply; the resulting errors are suppressed.

### DIFF
--- a/ingeniamotion/fsoe_master/handler.py
+++ b/ingeniamotion/fsoe_master/handler.py
@@ -223,7 +223,7 @@ class FSoEMasterHandler:
         While the master is still completing its initial handshake with the
         slave (``__in_initial_reset``), the slave may still be echoing stale
         data from a previous session (e.g. a frozen safety PDU) or replying
-        with all-zero or stale data before it is ready. This can trigger expected,
+        with all-zero data before it is ready. This can trigger expected,
         transient errors (e.g. ``RESET_STAY1``) that resolve on their own once
         the slave catches up. These are suppressed here instead of being
         forwarded, to avoid reporting spurious errors during startup.
--- a/ingeniamotion/fsoe_master/handler.py
+++ b/ingeniamotion/fsoe_master/handler.py
@@ -29,7 +29,6 @@ from ingeniamotion.fsoe_master.fsoe import (
     FSoEDictionaryItemOutput,
     State,
     StateData,
-    StateReset,
     calculate_sra_crc,
 )
 from ingeniamotion.fsoe_master.parameters import (
@@ -107,9 +106,14 @@ class FSoEMasterHandler:
         self.__state_is_data = threading.Event()
         self.__stopping = False
 
-        # The saco slave might take a while to answer with a valid command
-        # During it's initialization it will respond with 0's, that are ignored
-        # To avoid triggering additional errors
+        # Baseline slave reply established since start(), or None if none has
+        # been seen yet. A reply that matches it isn't a new answer - whether
+        # that's a frame left over from a previous session or the slave still
+        # booting (e.g. answering with 0's before it's ready). It's still fed
+        # to the master handler as normal; __in_initial_reset reflects, per
+        # reply, whether it matched this baseline - see set_reply() and
+        # __report_error_callback().
+        self.__last_slave_message: Optional[bytes] = None
         self.__in_initial_reset = False
 
         # Parameters that are part of the system
@@ -219,7 +223,7 @@ class FSoEMasterHandler:
         While the master is still completing its initial handshake with the
         slave (``__in_initial_reset``), the slave may still be echoing stale
         data from a previous session (e.g. a frozen safety PDU) or replying
-        with all-zero data before it is ready. This can trigger expected,
+        with all-zero or stale data before it is ready. This can trigger expected,
         transient errors (e.g. ``RESET_STAY1``) that resolve on their own once
         the slave catches up. These are suppressed here instead of being
         forwarded, to avoid reporting spurious errors during startup.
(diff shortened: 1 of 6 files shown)


ingeniamotion/fsoe_master/handler.py
```diff
@@ -219,7 +223,7 @@ def __report_error_callback(self, transition_name: str, error: str) -> None:
         While the master is still completing its initial handshake with the
         slave (``__in_initial_reset``), the slave may still be echoing stale
         data from a previous session (e.g. a frozen safety PDU) or replying
-        with all-zero data before it is ready. This can trigger expected,
+        with all-zero or stale data before it is ready. This can trigger expected,
```
julieta-prieto: comment repeated, `the slave may still be echoing stale data from a previous session (e.g. a frozen safety PDU)`, this is mentioned just one line above

ingeniamotion/fsoe_master/handler.py
```diff
@@ -480,6 +483,15 @@ def set_reply(self) -> None:
...
             return
 
+        # First reply since start() can't yet correspond to anything this
+        # session asked for, and there's nothing prior to compare it to
+        # either - treat it the same as a genuine repeat.
+        self.__in_initial_reset = (
+            self.__last_slave_message is None or reply == self.__last_slave_message
+        )
+        if self.__last_slave_message is None:
+            self.__last_slave_message = reply
```
julieta-prieto, line 493: If you take a look at https://github.com/ingeniamc/ingeniamotion/pull/618, you'll see that this was found when debugging:

```
REPLY_TRACE index=1 state=None reply=2a0000c42d0000
...
REPLY_TRACE index=50 state=<class 'fsoe_master.fsoe_master.StateReset'> reply=528800bca30100
REPLY_TRACE index=1 state=None reply=2a050025ae0000
REPLY_TRACE index=2 state=<class 'fsoe_master.fsoe_master.StateSession'> reply=2a050025ae0000
REPLY_TRACE index=3 state=<class 'fsoe_master.fsoe_master.StateSession'> reply=2a0000c42d0000
ERROR_TRACE state=<class 'fsoe_master.fsoe_master.StateSession'> reply_index=3 ... transition=SESSION_RESET1 ...
ERROR_TRACE state=<class 'fsoe_master.fsoe_master.StateSession'> reply_index=4 ... transition=SESSION_STAY2 error=Invalid slave CRC
```

In this case, the stale reply is `2a0000c42d0000` which was already seen in the previous exchange, but there is `2a050025ae0000` in the middle. How would this `last_slave_message` react to that? Will this ever be detected if there is a different message before the stale one?

tests/fsoe/test_handler.py
```diff
@@ -284,6 +286,351 @@ def mock_send_receive_processdata(*args, **kwargs):
...
+        FSoENetwork(handler, FSoESlave()).exchange_to_data()
+
+        assert handler.state == FSoEState.DATA
+        assert errors == []
+    finally:
+        handler.delete()
+
+
+@pytest.mark.fsoe
+def test_stale_data_reply_survives_master_replacement() -> None:
```
julieta-prieto: can we rename this one? it looks like if we're trying to demonstrate a bug when we're trying to demonstrate that the master correctly ignores that right?

julieta-prieto (commented):
Can you please remove the marker https://github.com/ingeniamc/ingeniamotion/blob/2f5e5cbe27ff30b914bd25b8b09c42dfe96a29af/tests/fsoe/test_fsoe.py#L238 so that the test is executed again?

ingeniamotion/fsoe_master/handler.py
polfeliu: as talked, the situation cannot be reproduced so for now we're not acting on this

julieta-prieto (approved):

MartinALL91 (approved):

PR #644, opened 2026-08-03
Ingm 780 fix test get actual position

julieta-prieto:
The failure for ECAT product was the same than in https://github.com/ingeniamc/ingeniamotion/pull/613, invalid CRC message from the absolute encoder with BiSS-C protocol.

This is a known issue that is supposed to be fixed in the upcoming 2.11.0 release, so it has been marked as `biss_c_flaky`, and will be skipped for fw <= 2.10.0. This test now runs in SIRIUS setup.

With 100 repeated runs, it still failed sometimes, but with an assertion error due to the strict comparison:
image
The captured arrays show the position is still changing slightly while those samples are collected. Because of that, a new waiting time has been introduced to wait for the drive to stabilize in the desired position. After that, the samples are taken and compared against the 0.5 threshold.

Changes tested on commit https://github.com/ingeniamc/ingeniamotion/pull/644/changes/74b245fc4bc2b64b1cbf84f3847227c3621e9068 with [this pipeline](http://jenkins.ingenia/job/Novanta%20Motion%20-%20Ingenia%20-%20Git/job/ingeniamotion/job/PR-644/14/stages/?selected-node=401) (link may expire):
* 80 tests pass with SSI protocol (expected - no CRC check)
* 80 tests pass with rc and BiSS-C protocol: upcoming release does fix the issue

### DIFF
--- a/tests/test_motion.py
+++ b/tests/test_motion.py
@@ -382,6 +382,7 @@ def test_ramp_generator(mocker, init_v, final_v, total_t, t, result):
 @pytest.mark.biss_c_flaky(
     "Sporadically fails on ABS BiSS-C config, will be skipped for certain firmware versions"
 )
+@pytest.mark.repeat(100)
 def test_get_actual_position(mc, alias, position_value):
     mc.motion.set_operation_mode(OperationMode.PROFILE_POSITION, servo=alias)
     mc.motion.motor_enable(servo=alias)
--- a/tests/test_motion.py
+++ b/tests/test_motion.py
@@ -379,8 +379,10 @@ def test_ramp_generator(mocker, init_v, final_v, total_t, t, result):
 @pytest.mark.soem
 @pytest.mark.canopen
 @pytest.mark.parametrize("position_value", [-4000, -1000, 1000, 4000])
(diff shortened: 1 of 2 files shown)


tests/test_motion.py
```diff
@@ -379,13 +379,36 @@ def test_ramp_generator(mocker, init_v, final_v, total_t, t, result):
...
 @pytest.mark.not_valid_for_product(part_number="EVE-*")
 def test_get_actual_position(mc, alias, position_value):
     mc.motion.set_operation_mode(OperationMode.PROFILE_POSITION, servo=alias)
     mc.motion.motor_enable(servo=alias)
     mc.motion.move_to_position(position_value, servo=alias, blocking=True, timeout=10)
+
+    # Wait for the position to stabilize
+    stability_timeout = 3
+    stability_interval = 0.1
+    stability_tolerance = 20
```
MartinALL91: I think this tolerance should be proportional to the encoder resolution instead of being hardcoded.

julieta-prieto: @MartinALL91 changed

MartinALL91 (approved):

GullinBustin (approved):

PR #645, opened 2026-08-03
MOT3-5457-feedbacks-wizard-avoid-transient-configuration-4-feedbacks-during-apply-rollback-internal-generator-swap

polfeliu:
This pull request introduces significant improvements to how feedback selector registers are managed in the feedback wizard tests, ensuring that the drive's maximum simultaneous feedback limit is never exceeded during configuration changes or restoration. It also adds comprehensive testing to validate these constraints and updates dependencies to support the new features. The api of the feedbacks has made the code and testing a bit convoluted. A improvement of the api and code is on its way: https://github.com/ingeniamc/ingeniamotion/pull/647/changes

**Feedback Configuration Management:**

* The feedback selector configuration logic in `Feedbacks.feedback_setting()` was refactored to stage and apply sensor assignments in an order that guarantees the drive's simultaneous feedback limit is never exceeded. This prevents transient invalid states during reconfiguration.
* A new `_restore_configuration` method was added to `Feedbacks`, ensuring feedback selector registers are restored to their baseline values in a safe order, again respecting the drive's feedback limit.
* The base test class now calls the appropriate `_restore_configuration` method after every test run, ensuring proper cleanup for all wizard tests.
* The context manager for drive operations is now exposed to allow access to baseline configuration during restoration.

**Testing Enhancements:**

* A new test, `test_feedback_test_respects_and_restores_the_drive_feedback_limit`, exhaustively checks that the wizard test never configures more than the allowed number of feedbacks at once and always restores the initial configuration, regardless of the starting state. It introduces utility helpers for tracking and generating feedback selector states. The test validates around 200 feedback combinations with subtests, although the feedback test has been patched to avoid running it, this takes about 2 minutes of hardware


pyproject.toml
```diff
@@ -117,6 +117,7 @@ pytest-mock = "==3.15.1"
 pytest-repeat = "==0.9.4"
 pytest-rerunfailures = "==15.1"
 pytest-console-scripts = "==1.4.1"
+pytest-subtests = "==0.15.0"
```
polfeliu, line 120: added this deprecateed dependency to be able to implement subtests. when we drop 3.9, we will be able to use pytest 9, that already contains that feature builtin

ingeniamotion/wizard_tests/feedbacks_tests/feedback_test.py
```diff
@@ -145,22 +146,75 @@ def __check_resolution(self, displacement: float) -> ResultType:
...
         """Set the feedback for the test.
 
         Raises:
             TestConfigurationError: If the feedback resolution is not greater than 0.
         """
-        # First set all feedback to feedback in test, so there won't be
-        # more than 5 feedback at the same time
-        self.mc.configuration.set_commutation_feedback(
-            self.sensor, servo=self.servo, axis=self.axis
+        feedback_slots = (
```
polfeliu, line 234: in my opinion this is code with this declaration here and getters and setters iteration is horrendus, but I didn't find another way.
Here you can see a preview of the idea of the refactor that I would like to do https://github.com/ingeniamc/ingeniamotion/pull/647/changes#diff-572170e083b43902c696c5ec8db72ea4bb054a83da91dc7a733d9890a25f340dL156

julieta-prieto (approved):

MartinALL91 (approved):

PR #647, opened 2026-08-03
INGM-801-feedbacks-refactor

polfeliu:
This pull request introduces a major refactor of how feedback sensors are managed and accessed for axes in the codebase as a follow-up the horrible code introduce on https://github.com/ingeniamc/ingeniamotion/pull/645.
The new `AxisFeedbacks` API provides a typed, slot-based interface for feedback configuration, replacing the previous flat `Feedbacks` class, which now is a small compatiblity seam. This change brings safer, more maintainable, and more expressive feedback management throughout the codebase, especially in test logic. Additionally, cached properties improve performance for expensive or stateful properties. The test infrastructure and feedback-related tests have been updated to use the new API, resulting in cleaner and more robust test code.

### Feedbacks API Refactor

* Introduced `Axis.feedbacks` (`AxisFeedbacks`), a typed API for managing an axis's five feedback slots, including per-sensor-type `Encoder` subclasses and immutable `FeedbacksConfiguration` snapshots. This replaces the flat `Feedbacks` class, which now delegates to the new API for backward compatibility.
* Added `AxisFeedbacks.set_configuration`, enabling safe transitions between feedback slot configurations with write ordering and temporary parking to respect hardware limits.
* Added validation to ensure only legal sensor types are assigned to feedback slots, based on drive dictionary constraints. 
* Updated all relevant feedback test classes and test setup logic to use the new `AxisFeedbacks` API for setting, reading, and updating feedback configurations, resolutions, and polarities.

### Property Caching and Code Cleanliness

* Used `cached_property` for instance-cached properties like `errors`, `feedbacks`, and accessors for axis and motion node, improving performance and code clarity. 
* Removed redundant instance attributes for errors and feedbacks, now built on first access. 
* Renamed the base feedback test class to `FeedbacksTest` and updated all references for clarity and correctness. 
### Test Infrastructure Updates

* Added cached property accessors in `BaseTest` for `_axis`, `_motion_node`, and `_axis_feedbacks`, centralizing and simplifying test access to the new feedback API. (
* Updated test classes and test setup/teardown logic to use the new feedback API, leading to more robust and readable test code.


ingeniamotion/feedbacks.py
```diff
@@ -1,55 +1,711 @@
...
+        Args:
+            slot: Feedback slot to update.
+            encoder: Encoder to assign to that slot.
+
+        Returns:
+            A new configuration with the slot updated.
+        """
+        return FeedbacksConfiguration({**self.__encoders, slot: encoder})
+
+    def replace(self, changes: Mapping[FeedbackSlot, Encoder]) -> "FeedbacksConfiguration":
```
julieta-prieto: what do we need this method for?

ingeniamotion/feedbacks.py
```diff
@@ -1,55 +1,711 @@
...
+        """
+        # Ensure the requested slot belongs to this configuration.
+        self.encoder_at(slot)
+        active_sensors = self.active_sensors()
+        return (
+            len(active_sensors) < MAX_SIMULTANEOUS_FEEDBACKS
+            or encoder.SENSOR_TYPE in active_sensors
+        )
+
+    def active_encoders_in_order(self) -> tuple[Encoder, ...]:
```
julieta-prieto, line 455: Do you think that this should be a method of this class?
It looks like a helper needed for `_find_feedback_order`

ingeniamotion/feedbacks.py
```diff
@@ -1,55 +1,711 @@
...
+        slot can be temporarily assigned an encoder that is already active in
+        another slot to free one of the drive's limited sensor entries. Keeping
+        the first-seen order makes that search deterministic.
+        """
+        return tuple(dict.fromkeys(self.__encoders.values()))
+
+
+def _find_feedback_order(  # noqa: C901
+    target: FeedbacksConfiguration,
+    state: FeedbacksConfiguration,
```
julieta-prieto: A bit confusing to call `state` the current configuration, I wouldn't have guessed without reading the docstring.
What do you think of renaming to `target_config`, `current_config`?

ingeniamotion/feedbacks.py
```diff
@@ -1,55 +1,711 @@
...
+    pending: tuple[FeedbackSlot, ...],
```
julieta-prieto, line 469: why is this sent an external argument?

If you have a target and a current state, is not possible to derive which one are pending within this method?

ingeniamotion/feedbacks.py
```diff
@@ -1,55 +1,711 @@
...
-    COMMUTATION_FEEDBACK_REGISTER = "COMMU_ANGLE_SENSOR"
-    REFERENCE_FEEDBACK_REGISTER = "COMMU_ANGLE_REF_SENSOR"
-    VELOCITY_FEEDBACK_REGISTER = "CL_VEL_FBK_SENSOR"
-    POSITION_FEEDBACK_REGISTER = "CL_POS_FBK_SENSOR"
-    AUXILIAR_FEEDBACK_REGISTER = "CL_AUX_FBK_SENSOR"
+    COMMUTATION_FEEDBACK_REGISTER = COMMUTATION_FEEDBACK_REGISTER
+    REFERENCE_FEEDBACK_REGISTER = REFERENCE_FEEDBACK_REGISTER
+    VELOCITY_FEEDBACK_REGISTER = VELOCITY_FEEDBACK_REGISTER
+    POSITION_FEEDBACK_REGISTER = POSITION_FEEDBACK_REGISTER
+    AUXILIAR_FEEDBACK_REGISTER = AUXILIARY_FEEDBACK_REGISTER
```
julieta-prieto, line 684: why do you need to redefine this here if they are already global?

ingeniamotion/feedbacks.py
```diff
@@ -680,7 +1200,7 @@ def get_feedback_polarity(
         feedback: SensorType,
         servo: str = DEFAULT_SERVO,
         axis: int = DEFAULT_AXIS,
-    ) -> Union[int, FeedbackPolarity]:
+    ) -> FeedbackPolarity:
```
julieta-prieto, line 1195: Don't you think that changing the return value of a public method can be dangerous in terms of breaking changes for users that may be using it?

ingeniamotion/feedbacks.py
julieta-prieto: I see that you use it in the tests. I am not a big fan of adding methods to classes just to fix the tests, is there any other use case apart from that?

tests/test_feedbacks.py
```diff
@@ -453,5 +431,227 @@ def test_instance_sensor_type(mc, alias):
         (SensorType.QEI2, "FBK_DIGENC2_POLARITY"),
     ],
 )
-def test_get_feedback_polarity_register_uid(mc, sensor, register):
-    assert mc.configuration.get_feedback_polarity_register_uid(sensor) == register
+def test_encoder_polarity_register_uid(axis, sensor, register):
```
julieta-prieto, line 434: docstring please (I know it is an easy one, but we agreed on adding them as good practice)

tests/test_feedbacks.py
```diff
@@ -453,5 +431,227 @@ def test_instance_sensor_type(mc, alias):
...
+    assert slot.get_encoder_type() == SensorType.HALLS
+
+
+@pytest.mark.virtual
+def test_set_encoder_type_rejects_unsupported_sensor(axis: "Axis") -> None:
+    """A slot rejects a sensor type the drive's dictionary does not allow for it."""
+    slot = axis.feedbacks.auxiliary
+    assert not slot.supports(SensorType.INTGEN)
+
+    with pytest.raises(ValueError) as exc_info:
```
julieta-prieto, line 459: I like to use `match` to check exception test:
```py
pytest.raises(
      ValueError,
      match=(f"INTGEN is not a valid sensor type for {AUXILIAR_FEEDBACK_REGISTER}."),
  )
```

tests/test_feedbacks.py
```diff
@@ -453,5 +431,227 @@ def test_instance_sensor_type(mc, alias):
...
+    encoder.set_polarity(FeedbackPolarity.REVERSED)
+    assert encoder.get_polarity() == FeedbackPolarity.REVERSED
+
+
+@pytest.mark.virtual
+def test_internal_generator_has_no_resolution_or_polarity(axis: "Axis") -> None:
+    """The internal generator rejects operations that require a physical encoder."""
+    encoder = axis.feedbacks.get_sensor(SensorType.INTGEN)
+
+    with pytest.raises(ValueError) as exc_info:
```
julieta-prieto, line 484: same (`match`)

tests/test_feedbacks.py
```diff
@@ -453,5 +431,227 @@ def test_instance_sensor_type(mc, alias):
...
+    axis: "Axis", sensor, register, raw_value, expected_resolution
+):
+    """Each concrete Encoder subclass computes its resolution from its own register."""
+    axis.write(register, raw_value)
+    encoder = axis.feedbacks.get_sensor(sensor)
+    assert encoder.get_resolution() == expected_resolution
+
+
+@pytest.mark.virtual
+def test_feedback_configuration_updates_are_immutable_and_ordered(axis: "Axis") -> None:
```
julieta-prieto: In which situation in our code (not tests) is a configuration copy needed?

tests/test_feedback_test.py
```diff
@@ -57,9 +57,14 @@ def test_bldc_feedback_setting_raises_on_zero_resolution(mc, alias):
         INCREMENTAL_ENCODER_1_RESOLUTION_REGISTER, 0, servo=alias, axis=axis
     )
     feedback_test = DigitalIncremental1Test(mc, alias, axis)
-    with pytest.raises(TestConfigurationError, match="resolution must be greater than 0"):
+    with pytest.raises(TestConfigurationError) as exc_info:
```
julieta-prieto, line 60: `with pytest.raises(TestConfigurationError,match="...")`?

tests/test_feedback_test.py
```diff
@@ -277,81 +304,234 @@ def _feedback_configurations(
...
+    seed = _feedback_transition_seed(setup_specifier)
+    selected_pair_indices = _feedback_transition_pair_indices(pair_count, seed, setup_specifier)
+
+    for pair_index in selected_pair_indices:
+        current_index, target_index = divmod(pair_index, len(configurations))
+        current = configurations[current_index]
+        target = configurations[target_index]
+        state = current
+
+        with subtests.test(
```
julieta-prieto, line 509: If any iteration (or several) fails, is this properly reported in the dashboard?

julieta-prieto: Maybe you can force a failure and see how it looks like

tests/test_feedbacks.py
polfeliu: match is a regex, so if you are adding trailing additional details this will not force you to update the test.

you would need to write this. To me it's not as readable:
`match=rf"^INTGEN is not a valid sensor type for {AUXILIAR_FEEDBACK_REGISTER}\.$"`

ingeniamotion/feedbacks.py
polfeliu: I agree that is a bit weird, that is why it has explanation, but self.__encoders is private so somehow you have to expose which active encoders are there. which would normally done by a set. But find feedback order relies on that order to be stable for the search algorithm, so... I'm not sure.

Do you have proposal?

ingeniamotion/feedbacks.py
polfeliu: note that this function is private. the public method is down below and doesn't have this argument

this argument only exists because ths search is recursive and it avoid recalculating the differences for every iteration step.

Also if we were to calculate it on each iteration step I'm not sure if the algorithm would recreate the same sequence each time, as sometimes there are multiple sequences possible to transition from one state to another

ingeniamotion/feedbacks.py
polfeliu: just api compatibility

ingeniamotion/feedbacks.py
polfeliu: i kind of agree on this one, but this api is very weird on my opinion, and is inconsistent with the rest of getter functions...

I could not find any other Union[int, SomeEnum] on the codebase.

I also di a gitblame and this was introduced on the first writing: https://github.com/ingeniamc/ingeniamotion/commit/d36d1f471cc7ca4de15d7a0753515568523823b2

Not as an a reaction to a bug, so it's possible this has never been the case.

Typing wise the change has no effect, because FeedbackPolarity is already a int because its IntEnum:
https://github.com/ingeniamc/ingeniamotion/blob/97c1b37a93e7f32f2658dd35959c7367bc5a7465/ingeniamotion/enums/__init__.py#L210-L211

the only change is the raising of valuerror on this weird case

ingeniamotion/feedbacks.py
```diff
@@ -1,55 +1,711 @@
...
+            if suffix is not None:
+                return [(slot, parking_encoder), *suffix]
+    return None
+
+
+class AxisFeedbacks:
+    """Class to manage the feedback slots of an axis.
+
+    Attributes:
+        commutation: The commutation feedback slot.
```
MartinALL91: These are properties, so they should not be included in the class-level `Attributes` section according to the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#s3.8.4-comments-in-classes). They are already documented in their respective docstrings.

CHANGELOG.md
```diff
@@ -1,13 +1,23 @@
 # Changelog
 
-## [0.10.4] - 2026-08-13
+## [Unreleased]
 ### Added
-- Add dynamic forced phasing test
 - Add SDCP servo support.
+- Add `Axis.feedbacks` (`AxisFeedbacks`), a typed API for an axis's five feedback slots: `FeedbackSlot` for reading/writing a slot's sensor type, per-sensor-type `Encoder` subclasses for resolution and polarity, and `FeedbacksConfiguration`, an immutable snapshot of the encoder assigned to every slot.
```
MartinALL91: Could we make this a bit shorter? I think the changelog should focus on the user-facing changes rather than the implementation details.

tests/test_feedbacks.py
julieta-prieto: why not

```python
import re

pytest.raises(
      ValueError,
      match=re.escape(f"INTGEN is not a valid sensor type for {AUXILIAR_FEEDBACK_REGISTER}."),
  )
```

I still think that it is much better than having an extra line to assert it

CHANGELOG.md
```diff
@@ -1,13 +1,23 @@
...
+- Add `AxisFeedbacks.set_configuration`, which safely transitions all feedback slots to a target configuration: writes are ordered, and a slot is temporarily parked on an already-active sensor when needed, so the drive's 4-simultaneous-feedback limit is never exceeded mid-transition.
+- Validate, before writing, that a sensor type is a legal value for a feedback slot's selector register according to the drive dictionary.
+
+### Changed
+- Feedbacks are now managed through `Axis.feedbacks` (`AxisFeedbacks`) instead of the flat `Feedbacks` class; `Feedbacks` remains available for backward compatibility and delegates to the new API.
+- Removed `get_feedback_polarity_register_uid`, `feedback_resolution_functions` and `get_sincos_encoder_resolution` from `Feedbacks`, and dropped SinCos support from `get_feedback_resolution`.
 
 ### Fixed
 - Feedback test symmetry check
```
MartinALL91: `Feedback test symmetry check` is part of the `0.10.4` release changelog, so it should remain under `0.10.4` rather than being moved to `Unreleased`.

tests/test_feedback_test.py
polfeliu: for the dashboard this is just one test. at collection time you don't know how many iterations it has

tests/test_feedbacks.py
polfeliu: same, you need the ^ and $ for the sequence to be completely matched

MartinALL91 (approved):

lucia-teruel (approved):

PR #659, opened 2026-08-10
Fix develop

polfeliu:
Skipping tests/setups that are broken.
- Motion tests with sirius @julieta-prieto is working on fixing
- STO abnormal on ethercat everest, created issue for fix the unreliable hw


tests/test_motion.py
```diff
@@ -285,6 +286,7 @@ def test_set_position(mc, alias, position_value):
     "Sporadically fails on ABS BiSS-C config, will be skipped for certain firmware versions"
 )
 @pytest.mark.not_valid_for_product(part_number="EVE-XCR-C")
+@pytest.mark.not_valid_for_product(part_number="EVS-NET-E", interfaces=[Interface.ECAT])
```
julieta-prieto: won't this skip sirius SSI which is working?
I would just set "2.11.0.005" to never

tests/test_motion.py
```diff
@@ -382,6 +384,7 @@ def test_ramp_generator(mocker, init_v, final_v, total_t, t, result):
     "Sporadically fails on ABS BiSS-C config, will be skipped for certain firmware versions"
 )
 @pytest.mark.not_valid_for_product(part_number="EVE-*")
+@pytest.mark.not_valid_for_product(part_number="EVS-NET-E", interfaces=[Interface.ECAT])
```
julieta-prieto: won't this skip sirius SSI which is working?

julieta-prieto (changes requested):

tests/test_motion.py
polfeliu: this is only for sirius, its still failing, I thought you were still working on that. See todays nightly http://jenkins.ingenia/job/Novanta%20Motion%20-%20Ingenia%20-%20Git/job/ingeniamotion/job/develop/2050/stages/?selected-node=405

julieta-prieto: @polfeliu what I say is that SIRIUS+BiSS-C (2.11.0.005) fails, SIRIUS+SSI passes.
If you skip the product, you are going to skip SSI as well. Why do you want to skip both?

julieta-prieto: image

This is the result from the Jenkinsfile you attached, there is no need to skip SSI tests.

In any case, please add the jira issue in the code, I don't think there is one created.

julieta-prieto (approved):

MartinALL91 (approved):

PR #678, opened 2026-08-26
Ingm 789 fix test emcy callback

julieta-prieto:
After running 100 iterations both on EVE-XCR-E 2.6.0 and 2.8.1, Wireshark captures conclude that it is not a software bug.

Out of 100 iterations, sometimes 5 or 7 fail, so there is still a chance that it will pass.

See [related issue](https://novantamotion.atlassian.net/browse/DRIVSUS-392) for the full issue investigation and results.

### DIFF
--- a/tests/test_communication.py
+++ b/tests/test_communication.py
@@ -809,9 +809,8 @@ def test_subscribe_register_updates(mc: "MotionController", alias: str) -> None:
 
 @pytest.mark.canopen
 @pytest.mark.soem
-@pytest.mark.not_valid_for_product(
-    part_number="EVE-XCR-E", xfail="https://novantamotion.atlassian.net/browse/DRIVSUS-392"
-)
+# https://novantamotion.atlassian.net/browse/DRIVSUS-392
+@pytest.mark.not_valid_for_product(part_number="EVE-XCR-E")
 def test_emcy_callback(mc: "MotionController", alias: str) -> None:
     emcy_test = EmcyTest()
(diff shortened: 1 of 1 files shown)


MartinALL91 (approved):

tests/test_communication.py
```diff
@@ -809,27 +809,41 @@ def test_subscribe_register_updates(mc, alias):
...
+        )
+        servo_alias, second_emcy = emcy_test.messages[1]
+        assert servo_alias == alias
+        assert second_emcy.error_code == 0x0000, (
+            f"Expected error code 0x0000, got {second_emcy.error_code:#06x}"
+        )
+        assert second_emcy.get_desc() == "No error", (
+            f"Expected description 'No error', got '{second_emcy.get_desc()}'"
+        )
+    finally:
```
GullinBustin, line 840: The "finally" part is the only change, right? Do you remember what the original problem was?

julieta-prieto: @GullinBustin yes, this was a possible added problem from this test. If any assertion failed, the code that is currently in the finally would not be executed.

I just changed that to make it safer, but it makes no difference with respect to the test behavior. There is still a bug that needs to be solved for the test to pass.

GullinBustin (approved):

tests/test_communication.py
polfeliu: dont think its necessary to change it here, but the set register is not needed.
the subscription to the emergencies is a bit leaky. not critical but created issue:
https://novantamotion.atlassian.net/browse/CIT-838

polfeliu (approved):

PR #683, opened 2026-08-27
Ingm 772 fix test homing on index pulse

julieta-prieto:
# Summary

`test_homing_on_index_pulse` was previously skipped for the affected products. When enabled, repeated executions could fail intermittently.

The homing status could already contain the Homing attained bit from a previous operation. In that case, checking only for an attained status could produce a false positive or make it unclear whether the new homing command had actually executed. The target-latch command also issued the clear and set writes consecutively without confirming that the drive had accepted either transition.

The initial-position fixture could additionally accumulate position between runs by moving relative to the current position, and the motor was not guaranteed to be disabled if setup failed.

Firmware versions EVE `2.7.0` through `2.8.1` are known to be invalid for this test and are explicitly excluded: see [related issue](https://novantamotion.atlassian.net/browse/COMOCOAPP-493) (fixed in 2.9.0).

See related PRs:
* https://github.com/ingeniamc/ingenialink-python/pull/911
* https://github.com/CeleraMotion/summit-testing-framework/pull/107 

# Changes
* Updated `Motion.target_latch()` to:
   * Clear the target-latch control bit.
   * Wait for the drive to acknowledge the cleared bit.
   * Set the bit again to generate a reliable rising edge.
   * Wait for the set state with a bounded timeout.
* Updated the homing status check to require a real transition:
   * Homing attained must first be cleared.
   * Homing attained must then be set again without a homing error.
* Added regression tests for stale attained status and the clear-to-attained transition.
* Made the initial-position fixture move to an absolute, deterministic position.
* Ensured the motor is disabled in a finally block.
* Added a polling interval so status checks do not busy-loop.
* Asserted the result of `power_cycle_and_restore()` before continuing error-queue tests.

# Validation
* Repeated the index-pulse homing test 100 times to expose intermittent behavior.
   * It always failed in some iteration for EVE-XCR-E 2.8.1.
   * Apparently the bug is not present in EVE-XCR-E 2.6.1 -> see [pipeline](http://jenkins.ingenia/job/Novanta%20Motion%20-%20Ingenia%20-%20Git/job/ingeniamotion/job/PR-683/22/stages/?selected-node=621) (link may expire).
       * FW team has not been able to confirm when this started happening.
* Over 2000 runs in total for CAP-XCR-E which was also skipped, the failure has not been reproduced.


ingeniamotion/motion.py
```diff
@@ -64,10 +67,30 @@ def target_latch(self, servo: str = DEFAULT_SERVO, axis: int = DEFAULT_AXIS) ->
...
+        control_word = self._wait_for_target_latch_bit(bit_set=False, servo=servo, axis=axis)
         new_control_word = control_word | self.CONTROL_WORD_TARGET_LATCH_BIT
         self.mc.communication.set_register(
             self.CONTROL_WORD_REGISTER, new_control_word, servo=servo, axis=axis
         )
+        self._wait_for_target_latch_bit(bit_set=True, servo=servo, axis=axis)
+
+    def _wait_for_target_latch_bit(self, bit_set: bool, servo: str, axis: int) -> int:
+        deadline = time.monotonic() + self.TARGET_LATCH_TIMEOUT_S
+        while True:
```
GullinBustin: Add the timeout check here instead of using while True.

pyproject.toml
```diff
@@ -119,9 +119,9 @@ pytest-rerunfailures = "==15.1"
 pytest-console-scripts = "==1.4.1"
 pytest-subtests = "==0.15.0"
 matplotlib = "==3.9.4"
-summit-testing-framework = {version = "0.3.1.post334+pr.103.b3"}
-ingenialink = {version = "7.6.2.post290+develop.be70ebf"}
-virtual_drive = {version = "0.1.1.post181+develop.a07ec8b"}
+summit-testing-framework = {version = "0.3.1.post392+pr.107.b23"}
+ingenialink = {version = "7.6.2.post354+pr.911.b25"}
+virtual_drive = {version = "0.1.1.post234+pr.116.b2"}
```
MartinALL91: I think we should use a development wheel instead, since PR-116 has already been merged.

ingeniamotion/motion.py
julieta-prieto: @GullinBustin changed

pyproject.toml
julieta-prieto: @MartinALL91  changed

MartinALL91 (approved):

GullinBustin (approved):

PR #687, opened 2026-08-31
Ingm 773 fix test homing on current position

julieta-prieto:
## Summary

Use the shared framework teardown behavior to prevent motion from leaking between tests in the homing test suite.
See [related PR](https://github.com/CeleraMotion/summit-testing-framework/pull/111) for context.

## Changes

- Updated the `summit-testing-framework` test dependency to the wheel containing the shared velocity-settling helper.
- Current-position homing cleanup:
  - Clears the target-latch bit.
  - Verifies that `DRV_STATE_CONTROL & 0x200 == 0`.
  - Verifies that the actual velocity settles after homing.
- No production `ingeniamotion` API changes were required.

## Tests

Tested with 100 runs [pipeline](http://jenkins.ingenia/job/Novanta%20Motion%20-%20Ingenia%20-%20Git/job/ingeniamotion/job/PR-687/40/stages/?selected-node=968) (link may expire).


ingeniamotion/homing.py
```diff
@@ -82,16 +82,17 @@ def homing_on_current_position(
...
+            self.mc.motion.set_operation_mode(OperationMode.HOMING, servo, axis)
 
-        # Perform the homing
-        self.mc.motion.target_latch(servo, axis)
-        # Restore op mode
-        self.mc.motion.set_operation_mode(prev_op_mode, servo, axis)
+            # Perform the homing
+            self.mc.motion.target_latch(servo, axis)
+        finally:
+            self.mc.motion.set_operation_mode(prev_op_mode, servo, axis)
```
polfeliu, line 95: is operation mode excluded in the context manager? should we create issue?

polfeliu (approved):

MartinALL91 (approved):

ingeniamotion/homing.py
julieta-prieto: @polfeliu I think it is already restored in the tests because the register is not excluded.

This change is in `homing.py` though. We added it here because `homing_on_current_position()` temporarily changes the drive’s operation mode to `HOMING` before calling `target_latch()`. If any configuration step or `target_latch()` raises an exception, the drive could otherwise remain in `HOMING`.

PR #688, opened 2026-09-01
Cit 841 check develop webhooks triggering with hardware builds

julieta-prieto:
Develop branch indexing/webhooks were using run policy tags to check whether or not hardware builds should run.

With the latest changes, the webhooks always pick the correct value from the last human/cron build.
This means that if the last build was a cron build, it will set nightly and weekend policy as true.

To fix this, instead of relying in policy tags, just ignore hardware builds for branch indexing or webhooks.


MartinALL91 (approved):

polfeliu (approved):
With this, a manually launched develop build will have the same configuration as a PR right?
This is something I have missed

julieta-prieto:
> With this, a manually launched develop build will have the same configuration as a PR right? This is something I have missed

A manually launched develop will follow the options that the user has selected, same as the PRs.

GullinBustin (approved):

lucia-teruel (approved):

PR #695, opened 2026-09-09
Ingm 776 fix test homing on switch limit timeout

julieta-prieto:
# Description
This PR makes the homing timeout test more reliable on EtherCAT and CANopen. It now waits for the motor to actually start and stop instead of relying on fixed delays and averaged readings.

It also makes sure the motor is disabled and the target latch is cleared when the test finishes or fails.

# Changes
* Fixed homing timeout test flakiness across EtherCAT and CANopen by polling instantaneous actual velocity for bounded motion-start and motion-stop detection.
* Added reliable cleanup that disables the motor and clears the target latch, including failure paths
   * `_cleanup_homing_motion` protective clause (it was already on `initial_position`).
* Added typing to all tests in `test_homing.py`

# Tests
* Tested with 100 runs in commit [b01af0d](https://github.com/ingeniamc/ingeniamotion/pull/695/commits/b01af0dfc6c1d95661b6fade926387ada676958e) with [pipeline](http://jenkins.ingenia/job/Novanta%20Motion%20-%20Ingenia%20-%20Git/job/ingeniamotion/job/PR-695/13/stages/?selected-node=329) (link may expire).


MartinALL91 (approved):

tests/test_homing.py
```diff
@@ -178,27 +237,31 @@ def test_homing_on_switch_limit_timeout(servo, mc, alias):
...
+            assert pytest.approx(0, abs=0.05) == mean_actual_velocity_position(
+                mc, alias, velocity=True
+            )
+            mc.motion.motor_enable(servo=alias)
+            mc.motion.target_latch(servo=alias)
+            _wait_for_homing_motion(mc, alias, homing_timeout / 1000)
+            time.sleep(homing_timeout / 1000)
+            _wait_for_homing_stop(mc, alias, homing_timeout / 1000)
+        finally:
+            _cleanup_homing_motion(mc, alias)
```
polfeliu, line 261: not sure if i already commetn this on anotehr pr. should we create issue to move this to framework?

I.E, any test that calls enable motor -> wait motor to stop at the end

julieta-prieto: @polfeliu it was already done here https://github.com/CeleraMotion/summit-testing-framework/pull/111

julieta-prieto: The framework teardown runs after the test body has finished. It cannot validate the homing timeout before the test result is decided.

Also, framework teardown does this:
* Disables the motor.
* Waits for zero velocity.
* Fault reset.

That wait confirms the motor stopped after teardown disabled it. It does not prove that the homing sequence stopped by itself after its timeout.

`_wait_for_homing_stop` runs inside the test, before cleanup, so it checks the actual homing behavior. Without it, the test could pass merely because teardown stopped the motor afterward.

So both have different jobs:

* `_wait_for_homing_stop`: verifies homing stopped as expected.
* `_cleanup_homing_motion`: clears the latch and performs immediate test cleanup.
* Framework teardown: final safety cleanup after the test.

GullinBustin (approved):

PR #696, opened 2026-09-09
MOT3-5669 Create object oriended monitoring module with check_version cache

polfeliu:
Introduces `MotionNodeCapture`, an object-oriented API that binds monitoring and disturbance operations to a `MotionNode`.

Monitoring version detection is now cached per motion node, avoiding repeated register reads while ensuring results affected by communication errors are not cached.

The existing `Capture` API remains available as a backward-compatible facade.

## Changes

- Add `MotionNode.capture` as a cached `MotionNodeCapture` instance.
- Move monitoring and disturbance status, enable, disable, and version detection operations to `MotionNodeCapture`.
- Cache successfully detected monitoring versions.
- Cache drives confirmed not to support monitoring and disturbance.
- Do not cache version detection results affected by communication errors.
- Preserve explicit version arguments without probing version registers.
- Preserve the existing `Capture` API and its IngeniaMotion exception contract.
- Read capture registers directly through the motion node servo.
- Add tests covering the object cache, version cache, communication failures, unsupported drives, explicit versions, and facade compatibility.

## Testing

- Added virtual-drive tests for monitoring version detection and caching.
- Added tests for monitoring and disturbance operations through both `MotionNodeCapture` and the existing `Capture` facade.

### DIFF
--- a/ingeniamotion/capture.py
+++ b/ingeniamotion/capture.py
@@ -26,6 +26,7 @@ from ingeniamotion.pdo import PDONetworkManager
 
 if TYPE_CHECKING:
     from ingeniamotion.motion_controller import MotionController
+    from ingeniamotion.motion_node import MotionNode
 
 
 class Capture:
@@ -285,6 +286,24 @@ class Capture:
             self.enable_disturbance(servo=servo)
         return disturbance
 
+    def _capture(
+        self, servo: str, version: Optional[MonitoringVersion] = None
+    ) -> "MotionNodeCapture":
+        """Return capture operations for a motion node and optional version."""
+        motion_node = self.mc._get_motion_node(servo)
+        if version is None:
+            return motion_node.capture
+
+        return MotionNodeCapture._with_version(motion_node, version)
+
+    def _resolve_version(
+        self, servo: str, version: Optional[MonitoringVersion]
+    ) -> MonitoringVersion:
+        """Return an explicit version or the bound capture's cached version."""
+        if version is not None:
+            return version
+        return self._capture(servo).version
+
     def _check_version(self, servo: str) -> MonitoringVersion:
         """Checks the version of the monitoring based on a given servo.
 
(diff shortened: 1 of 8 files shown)


MartinALL91 (approved):

ingeniamotion/capture.py
```diff
@@ -26,24 +28,245 @@
...
+        if not isinstance(value, int):
+            raise TypeError(type_error_message)
+        return value
 
 
 class Capture:
     """Capture."""
 
-    DISTURBANCE_STATUS_REGISTER = "DIST_STATUS"
+    DISTURBANCE_STATUS_REGISTER = MotionNodeCapture._DISTURBANCE_STATUS_REGISTER
```
julieta-prieto, line 255: why do you redefine them here instead of using the ones from `MotionNodeCapture`?

ingeniamotion/capture.py
```diff
@@ -285,6 +508,15 @@ def create_disturbance(
             self.enable_disturbance(servo=servo)
         return disturbance
 
+    def _capture(
+        self, servo: str, version: Optional[MonitoringVersion] = None
+    ) -> MotionNodeCapture:
+        """Return capture operations for a motion node and optional version."""
```
julieta-prieto, line 514: Arguments are completely missing from the docstring

tests/test_capture.py
```diff
@@ -357,31 +396,154 @@ def test_check_monitoring_version_v3(mc, alias):
...
+        nonlocal failed
+        if register == MotionNodeCapture._MONITORING_VERSION_REGISTER and not failed:
+            failed = True
+            raise ILError
+        return original_read(register, *args, **kwargs)
+
+    mocker.patch.object(motion_node.servo, "read", side_effect=read)
+
+    assert motion_node.capture.version == MonitoringVersion.MONITORING_V2
+    assert motion_node.capture.version == MonitoringVersion.MONITORING_V3
```
julieta-prieto, line 528: I don't get this, are you asserting that it is version 2 and then that it is version 3? 
How can it be both at the same time?

ingeniamotion/capture.py
```diff
@@ -26,24 +28,245 @@
...
+
+    def get_disturbance_status(self) -> int:
+        """Return the disturbance status of the motion node.
+
+        Returns:
+            Disturbance status.
+
+        Raises:
+            ILRegisterNotFoundError: If the register doesn't exist.
+            NotImplementedError: If the drive does not support monitoring and disturbance.
```
lucia-teruel: For consistency, could `get_monitoring_status()` also raise a `NotImplementedError` when Monitoring is not available?

lucia-teruel (approved):

ingeniamotion/capture.py
polfeliu: just api compatibility

tests/test_capture.py
polfeliu: read the patched read callback. failed acts as a a latch reset of the errror

ingeniamotion/capture.py
polfeliu: https://github.com/ingeniamc/ingeniamotion/pull/703

PR #709, opened 2026-09-16
Ingm 774 fix test commutation

julieta-prieto:
# Description
`test_commutation` was reported to fail in `CAP-XCR-E` setup.

To identify the issue, the test ran 100 times (several Jenkins iterations), and the following message appeared:
```sh
ABS1 disconnection or frame overlap in BiSS-C
```

This was already identified in https://github.com/ingeniamc/ingeniamotion/pull/613. Problems with ABS1 CRC checks are a known issue that is fixed in 2.11.0 release.

# Changes
* Updated `CAP-XCR-E` to `2.11.0`.
  * The newer dictionary exposes `FBK_DIGHALL_OFFSET_COMP_VALUE`, which can be modified during phasing, so it is treated as an expected changed register rather than a restore failure.
image

* Added exclusion marker for `CAP-XCR-E` in a certain firmware version range that is known to be flaky.

# Tests
* Verified that updating `CAP-XCR-E` to `2.11.0` the tests passed on 100 runs [pipeline](http://jenkins.ingenia/job/Novanta%20Motion%20-%20Ingenia%20-%20Git/job/ingeniamotion/job/PR-709/12/stages/) (link may expire) on commit [060c4e0](https://github.com/ingeniamc/ingeniamotion/pull/709/commits/060c4e0042432f269e22404d4717f978dd7e2261).


tests/test_all_drive_tests.py
```diff
@@ -37,6 +37,15 @@
 # Record stop opportunities for every wizard-test integration case in this module.
 pytestmark = pytest.mark.usefixtures("stoppable_trace_recorder")
 
+
+def _assert_commutation_starts_without_active_fault(mc, alias: str) -> None:
```
MartinALL91: Does this mean that we don't verify this for all tests?

julieta-prieto: @MartinALL91 we do a fault reset: https://github.com/CeleraMotion/summit-testing-framework/blob/b4cb13607fa47946f7187a4f32767fdd167c4d22/summit_testing_framework/connection/teardown_utils.py#L94.

We do not assert it though.

julieta-prieto: @MartinALL91 removed

MartinALL91 (approved):

tests/test_all_drive_tests.py
```diff
@@ -330,9 +331,11 @@ def test_secondary_ssi_test(
...
 @pytest.mark.soem
 @pytest.mark.canopen
-@pytest.mark.not_valid_for_specifier(
-    specifier="tests.setups.rack_specifiers.ECAT_SETUP@CAP-XCR-E",
-    skip_reason="https://novantamotion.atlassian.net/browse/INGM-774",
+@pytest.mark.not_valid_version_for_product(
+    part_number="CAP-*",
+    min="2.6.0",
+    max="2.10.0",
+    skip_reason="Flaky test for BISS-C configuration",
```
polfeliu, line 338: isn't the purpose of not_valid_version_for_product to indicate that the product has X failure with the issue?

the test is flaky because the product is flaky

tests/setups/rack_specifiers.py
```diff
@@ -226,15 +226,15 @@
...
                 extra_data={
                     __EXECUTION_POLICY_KEY: "always",
                     RANDOM_COMBINATIONS_SLICE_KEY: 0.1,
                     __TEST_CONFIGS_KEY: {
                         "ECAT_TEST_SESSIONS": PyTestConfig(
                             markers="soem",
-                            run_test_stage_uid="ethercat_capitan_2.9.0",
-                            stage_name="EtherCAT Capitan - FW. 2.9.0",
+                            run_test_stage_uid="ethercat_capitan_2.11.0",
+                            stage_name="EtherCAT Capitan - FW. 2.11.0",
```
polfeliu, line 237: im beggining to wonder if it makes sense to handroll this, we essentially have the info already... (out of scope)

tests/test_all_drive_tests.py
julieta-prieto: @polfeliu what do you mean?

The product is not flaky for firmware versions smaller than 2.6.0 according to what I have discussed with FW people.

I don't understand which change you expect me to do from your comment.

tests/setups/rack_specifiers.py
julieta-prieto: If it is not the firmware version, we would still need something since the stages must be unique, and there is another `ethercat_capitan` for version 2.4.0 if I am not mistaken.

Or are you proposing to remove those two arguments and derive them from existing information in CICD shared methods?

GullinBustin (approved):

